PR #8, opened 2021-12-03
Allow compiling on Debian 11

### LINKED ISSUE #7: Debian 11 - compile with mix
reproduce error:
```bash
cd onigumo
mix compile
```

error message: ** (Mix) You're trying to run :onigumo on Elixir v1.10.3 but it has declared in its mix.exs file it supports only Elixir ~> 1.12

Unfortunately last stable version for bullseye is 1.10.3 - https://packages.debian.org/search?suite=stable&section=all&arch=any&searchon=names&keywords=elixir

elixir ver. 1.12 is testing unstable - https://manpages.debian.org/bullseye/elixir/elixir.1.en.html

nappex:
Fix #7 

update configure information about allowed elixir version

### DIFF
--- a/mix.exs
+++ b/mix.exs
@@ -5,7 +5,7 @@ defmodule Onigumo.MixProject do
     [
       app: :onigumo,
       version: "0.1.0",
-      elixir: "~> 1.12",
+      elixir: "~> 1.10",
       start_permanent: Mix.env() == :prod,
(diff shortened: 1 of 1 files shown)


Glutexo:
I don’t really care about using the most recent Elixir version. :shipit:

PR #17, opened 2021-12-30
Add parentheses

nappex:
Elixir allow write functions without parentheses, but we'd like to write its because of education process.


Glutexo:
Elixir allows to write function calls without parentheses, but we’d like to put them in, for the sake of the education process.

Glutexo:
Dięx!

PR #20, opened 2021-12-31
Consolidate parentheses in calls

Glutexo:
* Use parentheses for function calls. That includes `import_config`,
  `config`, `test`, `assert`, `setup` or `defmock`.
* Do not use parentheses for [Kernel](https://hexdocs.pm/elixir/Kernel.html) calls and [special forms](https://hexdocs.pm/elixir/Kernel.SpecialForms.html). That includes `def`,
  `defp`, `defmodule`, `use` or `import`.

This makes the code consistent, yet avoiding strange syntax like `defmodule(…)`.

I am not sure what a good solution is. Maybe keeping asserts, test definitions and config calls without parentheses would be more readable. Yet I’d like to keep the number of exception as low as possible, sticking only to _Kernel_.

### DIFF
--- a/config/config.exs
+++ b/config/config.exs
@@ -1,4 +1,4 @@
-import(Config)
+import Config
 
 env = config_env()
 import_config("#{env}.exs")
--- a/config/test.exs
+++ b/config/test.exs
@@ -1,3 +1,3 @@
-import(Config)
+import Config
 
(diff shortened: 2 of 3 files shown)


nappex:
Ok why not. I'd prefer `assert` without parentheses but it just about how I know it from `python`. Unfortunately assert is really func in elixir.

nappex (approved):

PR #21, opened 2021-12-31
Remove unused doctest

Glutexo:
We do not have any test cases in our documentation strings. Thus `doctest` is not used and not needed. See [ExUnit.DocTest](https://hexdocs.pm/ex_unit/ExUnit.DocTest.html#doctest/2).


nappex:
Thanks it makes sense

PR #22, opened 2022-01-01
Format the code

nappex:
Elixir provides option `format` included in tool mix.
I format the code with command `mix format`


Glutexo:
Didn’t know about `mix format`. Thanks! It would be nice to have a pre-commit hook that runs that.

PR #34, opened 2022-01-21
Swap download/2 argument order

Glutexo:
The `http_client` argument is a configuration, `url` is an input. Keep
input values first to make it easier to use with `|>` piping.

```elixir
get_url() |> download(http_client)
```

### DIFF
--- a/lib/onigumo.ex
+++ b/lib/onigumo.ex
@@ -10,10 +10,10 @@ defmodule Onigumo do
     http = http_client()
 
     load_urls(@input_filename)
-    |> Enum.map(&download(http, &1))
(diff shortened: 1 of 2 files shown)


lib/onigumo.ex
```diff
@@ -10,10 +10,10 @@ defmodule Onigumo do
     http = http_client()
 
     load_urls(@input_filename)
-    |> Enum.map(&download(http, &1))
+    |> Enum.map(&download(&1, http))
```
Glutexo, line 13:
```suggestion
    |> Enum.map(fn url -> download(url, http) end)
```

nappex (approved):

PR #29, opened 2022-01-16
Test download of multiple URLs

nappex:
fix #26 

From @Glutexo: This required a slight change in the executive code so that the `download` function can accept a list of URLs without initializing the real environment application.

PR #40, opened 2022-01-22
Parametrize output filename

### LINKED ISSUE #32: Use :tmp_dir for all tests that touch filesystem
This may not be possible now because of lacking parametrization.

Glutexo:
This allows to use `tmp_dir` for `download/2` (now `/3`) tests.

Fixes #32 if not considering #29.

### DIFF
--- a/lib/onigumo.ex
+++ b/lib/onigumo.ex
@@ -10,16 +10,16 @@ defmodule Onigumo do
     http = http_client()
 
     load_urls(@input_filename)
-    |> Enum.map(&download(http, &1))
+    |> Enum.map(&download(http, &1, @output_filename))
   end
 
-  def download(http_client, url) do
+  def download(http_client, url, filename) do
     %HTTPoison.Response{
       status_code: 200,
       body: body
     } = http_client.get!(url)
 
-    File.write!(@output_filename, body)
+    File.write!(filename, body)
   end
 
   def load_urls(filepath) do
--- a/test/onigumo_test.exs
+++ b/test/onigumo_test.exs
@@ -8,7 +8,8 @@ defmodule OnigumoTest do
 
   setup(:verify_on_exit!)
 
-  test("download") do
+  @tag :tmp_dir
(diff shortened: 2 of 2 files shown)


nappex (approved):

lib/onigumo.ex
```diff
@@ -10,16 +10,16 @@ defmodule Onigumo do
     http = http_client()
 
     load_urls(@input_path)
-    |> Enum.map(&download(&1, http))
+    |> Enum.map(&download(&1, http, @output_path))
   end
 
-  def download(url, http) do
+  def download(url, http, path) do
```
Glutexo, line 16: Just thinking, wouldn’t it be more intuitive to have the literal input data collected on one side and the behaviors on the other side? Considering we already moved `url` to the beginning, I’d bring `path` there too.

```suggestion
  def download(url, path, http) do
```

lib/onigumo.ex
```diff
@@ -10,16 +10,16 @@ defmodule Onigumo do
     http = http_client()
 
     load_urls(@input_path)
-    |> Enum.map(&download(&1, http))
+    |> Enum.map(&download(&1, http, @output_path))
```
Glutexo, line 13:
```suggestion
    |> Enum.map(&download(&1, @output_path, http))
```

test/onigumo_test.exs
```diff
@@ -20,7 +21,7 @@ defmodule OnigumoTest do
       end
     )
 
-    assert(:ok == Onigumo.download(@url, HTTPoisonMock))
+    assert(:ok == Onigumo.download(@url, HTTPoisonMock, @output_path))
```
Glutexo, line 24:
```suggestion
    assert(:ok == Onigumo.download(@url, @output_path, HTTPoisonMock))
```

Glutexo (commented):
I realized we may have picked a suboptimal argument order for the parametrized filename. What do you think, @nappex?

PR #42, opened 2022-01-22
Make asserts more consistent and semantic

Glutexo:
Extracted function call results to variables. Kept literals only for simple expressions. Swapped the comparison sides order to be more semantic.


nappex (approved):

nappex:
@Glutexo  the conflict is resolved already.

Glutexo (commented):
✅

PR #41, opened 2022-01-22
Extract mock get! function

Glutexo:
For better readability and to be reused.

PR #43, opened 2022-01-22
Extract mock HTTP response body to a function

Glutexo:
DRY

### DIFF
--- a/test/onigumo_test.exs
+++ b/test/onigumo_test.exs
@@ -15,13 +15,13 @@ defmodule OnigumoTest do
       fn url ->
         %HTTPoison.Response{
           status_code: 200,
-          body: "Body from: #{url}"
+          body: body(url)
         }
       end
     )
 
     assert(:ok == Onigumo.download(HTTPoisonMock, @url))
-    assert("Body from: #{@url}" == File.read!(@filename))
+    assert(body(@url) == File.read!(@filename))
   end
 
 
@@ -35,4 +35,7 @@ defmodule OnigumoTest do
     assert(expected == Onigumo.load_urls(filepath))
   end
 
+  defp body(url) do
+    "Body from: #{url}"
+  end
 end


nappex:
My opinion is that this PR is unnecessary because of #41 . PR #41  is solving DRY of the body already.

Glutexo:
#41 only dries the mock `get!/1` function. The body composition is however a slightly different thing. The `get!/1` function is only one place from where it is used. The other place are the body asserts in the tests, where the composition logic is repeated.

In

Then in the tests, the expected body can be generated by

```elixir
assert(body(url) == File.read!(path))
```

instead of

```elixir
assert("Body from: #{url}\n" == File.read!(path))
```

Glutexo:
Merged [_elixir_](https://github.com/Glutexo/onigumo/tree/elixir) into this, so the change now becomes more apparent.

nappex:
> #41 only dries the mock `get!/1` function. The body composition is however a slightly different thing. The `get!/1` function is only one place from where it is used. The other place are the body asserts in the tests, where the composition logic is repeated.
> 
> In the `get/1` function, the body can be generated by
> 
> ```elixir
> %HTTPoison.Response{
>   status_code: 200,
>   body: body(url)
> }
> ```
> 
> Then in the tests, the expected body can be generated by
> 
> ```elixir
> assert(body(url) == File.read!(path))
> ```
> 
> instead of
> 
> ```elixir
> assert("Body from: #{url}\n" == File.read!(path))
> ```
> 
> .

Ok I've got it.

Glutexo:

nappex (approved):

test/onigumo_test.exs
```diff
@@ -34,7 +35,11 @@ defmodule OnigumoTest do
   defp get!(url) do
     %HTTPoison.Response{
       status_code: 200,
-      body: "Body from: #{url}\n"
+      body: body(url)
     }
   end
+
+  defp body(url) do
+    "Body from: #{url}"
```
Glutexo, line 43: This was my mistake when merging. 😞 It should have already contain the newline.

```suggestion
    "Body from: #{url}\n"
```

Glutexo (commented):
Spotted post-mortem mistake. 😞

test/onigumo_test.exs
Glutexo: Fixed in https://github.com/Glutexo/onigumo/commit/35cce947a85f99114abad41b4314c08a61f42068.

Glutexo:
Fixed in https://github.com/Glutexo/onigumo/commit/35cce947a85f99114abad41b4314c08a61f42068 directly pushed to master.

Glutexo:
Fixed in https://github.com/Glutexo/onigumo/commit/35cce947a85f99114abad41b4314c08a61f42068 directly pushed to [_elixir_](https://github.com/Glutexo/onigumo/tree/elixir).

Issue #19, opened 2021-12-30
Separate download func to smaller units


nappex:
We'd like to divide Onigumo.download to smaller parts with philosophy one func make one task.
This approach is more suitable for deep testing also. We'd like to use tmp_dir in test for testing files, for this intention we need to have the way how to give filepath as argument. When we reach it, we will be able to give tmp_dir as argument to make integration tests.
Onigumo.download make three different tasks at this moment.

Three tasks are:
1. make request
2. check request and match body of request to variable
3. write body content to file

Glutexo:
Cool!

Glutexo:
Partially addressed by #29.

Glutexo:
Streams would help this a lot. See #70.

Glutexo:
Partially addressed by #72. I am not sure yet how to split out the body extraction from the response object.

Glutexo:
#72 eventually resolved this completely.

PR #44, opened 2022-01-22
Extract URL input file composition

Glutexo:
DRY the URL input file composition to a new function. This is going to be useful when testing multiple URLs. Also allows to keep only one list of URLs around in the existing tests.

### DIFF
--- a/test/onigumo_test.exs
+++ b/test/onigumo_test.exs
@@ -63,7 +63,7 @@ defmodule OnigumoTest do
     input_urls = Enum.slice(@urls, 0, 1)
 
     path = Path.join(tmp_dir, @input_path)
-    content = urls_input(input_urls)
+    content = prepare_input(input_urls)
     File.write!(path, content)
 
     loaded_urls = Onigumo.load_urls(path)
@@ -87,7 +87,7 @@ defmodule OnigumoTest do
     }
   end
 
-  defp urls_input(urls) do
+  defp prepare_input(urls) do
     Enum.map(urls, &(&1 <> "\n"))
     |> Enum.join()
   end
--- a/test/onigumo_test.exs
+++ b/test/onigumo_test.exs
@@ -27,12 +27,17 @@ defmodule OnigumoTest do
 
   @tag :tmp_dir
   test("load URL from file", %{tmp_dir: tmp_dir}) do
+    urls = [@url]
+
     filepath = Path.join(tmp_dir, @testfile_with_urls)
-    content = @url <> " \n"
+    content = urls_input(urls)
     File.write!(filepath, content)
 
-    expected = [@url]
-    assert(expected == Onigumo.load_urls(filepath))
+    assert(urls == Onigumo.load_urls(filepath))
   end
 
+  defp urls_input(urls) do
+    Enum.map(urls, &(&1 <> " \n"))
+    |> Enum.join()
+  end
 end


Glutexo:

Glutexo:
Merged current [_elixir_](https://github.com/Glutexo/onigumo/tree/elixir) and resolved conflicts again.

test/onigumo_test.exs
```diff
@@ -39,6 +41,11 @@ defmodule OnigumoTest do
     }
   end
 
+  defp urls_input(urls) do
```
nappex, line 44:
```suggestion
  defp prepare_content(urls) do
```

test/onigumo_test.exs
```diff
@@ -24,12 +24,14 @@ defmodule OnigumoTest do
 
   @tag :tmp_dir
   test("load URL from file", %{tmp_dir: tmp_dir}) do
+    input_urls = [@url]
+
     path = Path.join(tmp_dir, @input_path)
-    content = @url <> "\n"
+    content = urls_input(input_urls)
```
nappex, line 30:
```suggestion
    content = prepare_content(input_urls)
```

test/onigumo_test.exs
```diff
@@ -24,12 +24,14 @@ defmodule OnigumoTest do
...
     File.write!(path, content)
 
-    urls = Onigumo.load_urls(path)
-    assert(urls == [@url])
+    loaded_urls = Onigumo.load_urls(path)
+    assert(loaded_urls == input_urls)
```
nappex, line 34:
```suggestion
    result = Onigumo.load_urls(path)
    assert(result == [@url])
```

test/onigumo_test.exs
```diff
@@ -24,12 +24,14 @@ defmodule OnigumoTest do
 
   @tag :tmp_dir
   test("load URL from file", %{tmp_dir: tmp_dir}) do
+    input_urls = [@url]
```
nappex, line 27:
```suggestion
```

nappex (commented):
This PR seems to me little bit overthinked and maybe unnecessary. I make some changes which make more sense. We can discuss it.

Glutexo:
Merged current [elixir](https://github.com/Glutexo/onigumo/tree/elixir) and resolved conflicts again.

Glutexo:
@nappex Is it? It extracts a piece of logic used to prepare an input file for the `load_urls` function. Makes converting URLs to `\n` terminated lines and joining them to a multi-line text file to appear in the tests only once. What makes you feel this is too overthought?

As for the naming, I used `urls_input`, because it is actually an input for the downloader and it contains URLs. _content_ sounds more vague to me, I’d even say it feels more like some kind of received payload. The _prepare_ part is good though. For tests, “to prepare inputs” is exactly what’s being done. 👍🏻 `prepare_input` may be a nice combination.

test/onigumo_test.exs
Glutexo: Commented in https://github.com/Glutexo/onigumo/pull/44#issuecomment-1029071593.

test/onigumo_test.exs
Glutexo: This suggestion is probably no longer valid after merging with current _elixir_. There is the `input_urls` variable containing what originally was in `[@url]`. I used `loaded_url` for symmetry, but `result` is good too. 👍🏻 Please revisit.

test/onigumo_test.exs
Glutexo: Commented in https://github.com/Glutexo/onigumo/pull/44#issuecomment-1029071593.

test/onigumo_test.exs
Glutexo: This is no longer valid, because `@url` ceased to exist and `@urls` contain more than one entry. I went for `Enum.slice(@urls, 0, 1)` to pick the first one.

test/onigumo_test.exs
nappex: ok, I've got it.

nappex:
> @nappex Is it? It extracts a piece of logic used to prepare an input file for the `load_urls` function. Makes converting URLs to `\n` terminated lines and joining them to a multi-line text file to appear in the tests only once. What makes you feel this is too overthought? Thanks for your input!
> 
> As for the naming, I used `urls_input`, because it is actually an input for the downloader and it contains URLs. _content_ sounds more vague to me, I’d even say it feels more like some kind of received payload. The _prepare_ part is good though. For tests, “to prepare inputs” is exactly what’s being done. 👍🏻 `prepare_input` may be a nice combination.

I feel really unreadable the combination:
```elixir
    input_urls = Enum.slice(@urls, 0, 1)

    path = Path.join(tmp_dir, @input_path)
    content = urls_input(input_urls)
```
Where is urls_input(input_urls). Do you see it? The names are used very unhappily in my view. Each of it is different but the names are almost same. I supposed that you will knit each other in some time. So I propose to setup more obvious names.
And thats why I'got feeling that could be overthought. At this point we are creating function which will be used just in two places. Yes, it will be nicer but is it really necessary to create this func? I am ok to create the func but I really appreciate if we think again about the names of these.

test/onigumo_test.exs
```diff
@@ -60,14 +60,14 @@ defmodule OnigumoTest do
 
   @tag :tmp_dir
   test("load a single URL from a file", %{tmp_dir: tmp_dir}) do
-    url = Enum.at(@urls, 0)
+    input_urls = Enum.slice(@urls, 0, 1)
```
nappex, line 63: Hmm clever usage. I've learnt somethign new again.

Glutexo:
The usage count may even drop if we go for parametrized tests as suggested in #59. Yet even if it end up being used only in one place, still it makes some sense to me to be extracted to a helper function. I mean, the tests should focus on the scenario, tell a story, not necessarily include technical implementation details. Moreover, the tests are still code and here #19 suggests splitting a single function in the executive part to even more atomic units. 🤷‍♂️

nappex (approved):

PR #45, opened 2022-01-22
Move input filename to config

### LINKED ISSUE #31: Extract input file name to Config
Use Application config to store the input URL file name. Like that it can be shared between the executive code and tests as well as overridden for the test environment.

Glutexo:
An input filename is a configuration value, it does not have to be a module constant. This extraction allows to use this value in tests too.

With this change, I removed the `@testfile_with_urls` attribute from the tests. Its purpose was to expose an otherwise inaccessible value in the `Onigumo` module. With `@input_file` moved to the config, this is no longer necessary, as it’s accessible with `Application.get_env(:onigumo, :input_file)`.

`load_urls/1` takes the filename as an argument and did not have to use the global value even before this change.

Fixes #31.

### DIFF
--- a/test/onigumo_test.exs
+++ b/test/onigumo_test.exs
@@ -4,7 +4,6 @@ defmodule OnigumoTest do
 
   @url "http://onigumo.org/hello.html"
   @filename "body.html"
-  @testfile_with_urls "urls.txt"
 
   setup(:verify_on_exit!)
 
@@ -27,7 +26,7 @@ defmodule OnigumoTest do
 
   @tag :tmp_dir
   test("load URL from file", %{tmp_dir: tmp_dir}) do
-    filepath = Path.join(tmp_dir, @testfile_with_urls)
+    filepath = Path.join(tmp_dir, "urls.txt")
     content = @url <> " \n"
     File.write!(filepath, content)
 
--- a/config/config.exs
+++ b/config/config.exs
@@ -1,4 +1,6 @@
 import Config
 
+config(:onigumo, :input_filename, "urls.txt")
+
 env = config_env()
 import_config("#{env}.exs")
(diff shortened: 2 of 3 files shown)


Glutexo:

I introduced a new change from the previous version. Instead of using a hardcoded `"urls.txt"` value in the tests, I load the value from the Application configuration. Even though in the tests it does not matter what the path actually is, the Application environment provide a sane value and prevents us from having hardcoded ones.

Glutexo:
Merged current [_elixir_](https://github.com/Glutexo/onigumo/tree/elixir) into this. Resolved conflicts, extended the usage of the config variable to the new tests.

Glutexo:
Merged current [_elixir_](https://github.com/Glutexo/onigumo/tree/elixir) into this. Resolved conflicts, extended the usage of the config variable to the new tests again.

Nice new piping has emerged.

```elixir
Application.get_env(:onigumo, :input_path)
|> load_urls()
|> download(http, path)
```

test/onigumo_test.exs
```diff
@@ -44,16 +43,18 @@ defmodule OnigumoTest do
   test("download URLs from the input file", %{tmp_dir: tmp_dir}) do
     expect(HTTPoisonMock, :get!, length(@urls), &get!/1)
 
+    env_path = Application.get_env(:onigumo, :input_path)
+    input_path = Path.join(tmp_dir, env_path)
```
nappex, line 47:
```suggestion
tmp_path = Application.get_env(:onigumo, :input_path) |> Path.join(tmp_dir, &1)
```

Unfortunately it is not working

nappex (approved):

test/onigumo_test.exs
Glutexo: How come the tests pass then?

Glutexo: Oh. I see. You mean that the suggested more concise approach with `&1` does not work. That’s unfornutately true. `&` captures can be only used in anonymous functions, not to override argument order when piping. Nice try though. 😃

Issue #11, opened 2021-12-11
Hash the file name


Glutexo:
So file names are unique and based on the URL. MD5?

Glutexo:
Slightly addressed by #49.

Glutexo:
This is slightly [getting](https://github.com/Glutexo/onigumo/pull/50#issuecomment-1024919004) complicated.

Glutexo:
When there is [`Base.url_encode64!/2`](https://hexdocs.pm/elixir/Base.html#url_decode64!/2), is there still a need to hash the URL and not use a reversible function instead? The reason here is that in the future, the input may not be only a URL, but some kind of normalized structure that would identify the request. There is no need to reconstruct the original data, only to recognize that is has been already downloaded. A hash has a constant length, with a limited character serves well as a filename and does not unnecessarily expose the internal structure of the input.

PR #49, opened 2022-01-22
Index output files

Glutexo:
Prepend a numbered index to the output files. This allows to download multiple URLs without overwriting the same file.

This would make much more sense if combined with #40. Like that it would be possible to create a filename generator function that could then be nicely replaced by a solution for #11.

Just proposing the idea for now as it would help tests in #29.

### DIFF
--- a/lib/onigumo.ex
+++ b/lib/onigumo.ex
@@ -10,16 +10,20 @@ defmodule Onigumo do
     http = http_client()
 
     load_urls(@input_filename)
-    |> Enum.map(&download(http, &1))
+    |> Enum.with_index()
+    |> Enum.map(fn {url, index} ->
+      download(http, url, index)
+    end)
   end
 
-  def download(http_client, url) do
+  def download(http_client, url, index) do
     %HTTPoison.Response{
       status_code: 200,
       body: body
     } = http_client.get!(url)
 
-    File.write!(@output_filename, body)
+    filename = "#{index}_#{@output_filename}"
+    File.write!(filename, body)
   end
 
   def load_urls(filepath) do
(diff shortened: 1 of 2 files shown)


Glutexo:
This suffers from the same problem as #50. Because of #40, the filename generation now needs to happen outside of the `download/3` function. That resulted on this functionality becoming untested as this now happens in the untested `main/0` function. Here too a filename generator and a new `download` definition may be the way to go.

A difference here is that this should change a file name, but not a path. It originally only expected a filename to be passed it to the `download/3` function. In the current stat if a full path is given, the indexation would cause improper behavior.

Glutexo:
Merged with current [_elixir_](https://github.com/Glutexo/onigumo/tree/elixir). Still not working with `:tmp_dir`.

Glutexo:
I fixed both the code and the tests. I am not very satisfied with the result.

Glutexo:
[Merged](https://github.com/Glutexo/onigumo/pull/49/commits/ecc4900c03bbaca31f233c615f7d27b7e2f6e42a) current [_elixir_](https://github.com/Glutexo/onigumo/tree/index) and resolved conflicts.

Glutexo:
After discussion with @nappex we agreed that indexing the output may not be the best solution, even though it is only temporary until proper hashing (#11, #50) is implemented. The main reason is that the result of the filename generator should remain equal for a given input, here, a URL. For indexes this is not true as they depend on the order in the input file.

A better solution here would be convert the URL to anything that’d be file-name safe. Strip special characters, or use some encoding like Base64. `Base.url_encode64!/2` produces URL and ** filename safe** strings, always the same for an unchanged URL.

Glutexo:
Merged current [_elixir_](https://github.com/Glutexo/onigumo/tree/elixir) and resolved conflicts.

Glutexo:
Closing in favor of #73.

PR #50, opened 2022-01-22
✨ Hash the filename using MD5

### LINKED ISSUE #11: Hash the file name

Glutexo:
Fixes #11. Invalidates #49.

Used an obsolete MD5 algorithm to produce file names from the URLs. The MD5 digest is nicely short and should be unique enough to suite our needs. There is no need for security here.

Created a new Hash module to avoid introducing new dependencies like [_ExCrypto_](https://hexdocs.pm/ex_crypto/ExCrypto.html), which doesn’t support MD5 nevertheless. This Elixir module uses and encapsulates OTP [`:crypto`](https://www.erlang.org/doc/man/crypto.html) module.

Introduced parametrized tests using for loops on the outside aided by `unquote` to access values outside the test function in the module definition.


Glutexo:
@nappex This has a lot of interesting points to discuss:

- Is really MD5 a good idea even for this simple non-cryptographic usage?
- Is it worth trying to avoid OTP modules? Is the `Hash` module enough encapsulation and how is that different from e.g. `HTTPoison` which too is a wrapper around OTP [`httpc`](https://www.erlang.org/doc/man/httpc.html).
- Wouldn’t it be worthy to publish the Hash module as a hex module and add it as a dependency?
- Are the parametrized tests a good practice, using `unquote`? How does that work?

test/hash_test.exs
```diff
@@ -0,0 +1,87 @@
...
+      <<61, 132, 37, 182, 234, 46, 254, 15, 167, 128, 117, 73, 44, 113,
+      159, 254>>
+    },
+    {
+      :hex,
+      "3d8425b6ea2efe0fa78075492c719ffe"
+    }
+  ]
+
+  @data "onigumo"
```
Glutexo, line 39: The best variable name ever. 😆

Glutexo:
Merged current [_elixir_](https://github.com/Glutexo/onigumo/tree/elixir) into this branch.

Implementing #40 caused that the current tests pass even with this new code in place. That however means that the new behavior is now untested. The [hashing](https://github.com/Glutexo/onigumo/blob/784078c91e622c04842ebf23fbfa8b6472621189/lib/onigumo.ex#L29) function is now called from [main](https://github.com/Glutexo/onigumo/blob/784078c91e622c04842ebf23fbfa8b6472621189/lib/onigumo.ex#L7), which too is untested.

There are only [tests](https://github.com/Glutexo/onigumo/blob/784078c91e622c04842ebf23fbfa8b6472621189/test/hash_test.exs) for the new [_Hash_](https://github.com/Glutexo/onigumo/blob/784078c91e622c04842ebf23fbfa8b6472621189/lib/hash.ex) module.

My current ideas are:

- [ ] Make a behavior for [_Hash_](https://github.com/Glutexo/onigumo/blob/784078c91e622c04842ebf23fbfa8b6472621189/lib/hash.ex) so it can be mocked.
- [ ] Create a version of [_download_](https://github.com/Glutexo/onigumo/blob/784078c91e622c04842ebf23fbfa8b6472621189/lib/onigumo.ex#L15) that would accept a filename function instead of a path.

Glutexo:
Merged current [_elixir_](https://github.com/Glutexo/onigumo/tree/elixir). The tests are now failing. The implementation was too naïve and did not distinguish between a file name and path. Because of that, temporary directories in tests can’t be used.

Glutexo:
Now with #64 having been merged, it may be possible to make this actually work.

Glutexo:
Resolved conflicts, fixed tests. Ready for review.

nappex:
I can not approve it because I dont understand the `unquote` I've found this article https://elixir-lang.org/getting-started/meta/quote-and-unquote.html. When I've tried `unquote` in `iex` I´ve got an error that unquote has to be used within `quote`, so How can you use it without `quoue` and tests are working MAGIC!! for me :/

Glutexo:
Thanks for the conflict resolution and the patch. I’ll come back to you with the (un)quoting later. 🙇🏻‍♂️

test/onigumo_test.exs
```diff
@@ -118,7 +115,7 @@ defmodule OnigumoTest do
   end
 
   defp assert_downloaded(url, tmp_dir) do
-    file_name = Base.url_encode64(url, padding: false)
+    file_name = Onigumo.create_file_name(url)
```
Glutexo, line 118: This change is slighly over the scope of this pull request. Before the change, the hashing method was used directly, duplicating the _create_file_name_ test. Now, _create_file_name_ becomes trusted and we don’t verify its result in this test.

nappex (approved):

PR #46, opened 2022-01-22
Remove http_client function

Glutexo:
There is no need to have a separate http_client function that only simply wraps around `Application.get_env/3` called from a single place. In the tests we are mocking _HTTPoison_ by replacing the configuration (environment) value, not by replacing this function. The mocking this continues to work even without a dedicated function.

This also removes the need of having a variable of a different name that would store the HTTP client module and moves the environment initialization to the top of the `main/0` function.

This pull request invalidates https://github.com/Glutexo/onigumo/pull/38.

PR #53, opened 2022-01-24
Always use env HTTP client

Glutexo:
`Application.get_env(:onigumo, :http_client)` returns `HTTPoison` in dev environment. This value can be used to call `start` upon. Doing so the dichotomy between using literal `HTTPoison` for `start`, but the env one for `get!` would disappear.

Related to https://github.com/Glutexo/onigumo/pull/46.

### DIFF
--- a/lib/onigumo.ex
+++ b/lib/onigumo.ex
@@ -6,8 +6,8 @@ defmodule Onigumo do
   @output_filename "body.html"
 
   def main() do
-    HTTPoison.start()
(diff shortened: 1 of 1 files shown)


Glutexo:
@nappex This is a simple change unrelated to #29 and it’s mergeable without conflicts.

nappex (approved):

Glutexo:
Thanks for the ✅! Merging.

PR #57, opened 2022-02-05
Consolidate instruction order in tests

Glutexo:
So that the order of argument initialization match the left to right order of the arguments in the following asserts. Taken from my [review](https://github.com/Glutexo/onigumo/pull/29#pullrequestreview-870324085) of #29.


nappex (approved):
yes, good one

PR #58, opened 2022-02-05
Replace variable with a pipe

Glutexo:


nappex (approved):
I supposed more elixiric way, why not

Glutexo:
elixiric 😃

PR #64, opened 2022-02-11
Only pass a dir path to download

Glutexo:
Using a static file path as an argument to the function that download more than one URL makes it impossible to generate the file names dynamically. Only kept the file path argument for the final `download/3` function accepting only one URL. Turned the argument of all other functions to a directory path, defaulting to the current working directory.

The new `download/1` function is untested, because there is no way to test it without custom behaviors or the `File` module mocking.

### DIFF
--- a/lib/onigumo.ex
+++ b/lib/onigumo.ex
@@ -11,9 +11,7 @@ defmodule Onigumo do
     download(http_client)
   end
 
-  def download(http_client, dir_path \\ nil)
-
-  def download(http_client, nil) do
+  def download(http_client) do
     dir_path = File.cwd!()
     download(http_client, dir_path)
   end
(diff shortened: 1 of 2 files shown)


Glutexo:
Just as with #49, I am not very happy with this and even less with another layer of the `download` functions. I did not want to change the architecture though, keeping the minimal amount of changes possible. It would be nice to refactor the function stack, but it can be done after this is merged just as well as before that.

Glutexo:
Merged current [_elixir_](https://github.com/Glutexo/onigumo/tree/elixir).

Glutexo:
Moved variables in tests around to improve legibility.

nappex (approved):

PR #65, opened 2022-02-11
Use default argument for load_urls

Glutexo:
Keep the logic of load_urls input in this function. Leverage default argument values to achieve this. As a result, the `download/2` function is simplified.

### DIFF
--- a/lib/onigumo.ex
+++ b/lib/onigumo.ex
@@ -12,8 +12,7 @@ defmodule Onigumo do
   end
 
   def download(http_client, path) do
-    Application.get_env(:onigumo, :input_path)
-    |> load_urls()
+    load_urls()
     |> download(http_client, path)
   end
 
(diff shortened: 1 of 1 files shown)


Glutexo:
[Merged](https://github.com/Glutexo/onigumo/pull/65/commits/63892e3b1354473e20489c46d0771928a9443bac) current [_elixir_](https://github.com/Glutexo/onigumo/tree/list_urls) and resolved conflicts.

lib/onigumo.ex
```diff
@@ -32,6 +31,13 @@ defmodule Onigumo do
     File.write!(file_path, body)
   end
 
+  def load_urls(path \\ nil)
+
```
nappex, line 35: Hmm I am confused what does it mean. Do you have some material or docs where did you find this approach, please?

Glutexo:
Closing without merging. I don’t want to do this any more. Feel free to comment or reopen.

PR #66, opened 2022-02-12
Convert load_urls to a Stream

Glutexo:
The downloading should work as a stream, everything from picking the URL from the file through the download from the Internet to the local drive file write should happen separately for every item, not by repeated enumeration over a list. ~~This required another version of the `download/3` function accepting a `Stream` and returning a `Stream`.~~ Started with converting the `load_urls/1` function to a `Stream`.

### DIFF
--- a/lib/onigumo.ex
+++ b/lib/onigumo.ex
@@ -15,6 +15,11 @@ defmodule Onigumo do
     Application.get_env(:onigumo, :input_path)
     |> load_urls()
     |> download(http_client, path)
+    |> Stream.run()
+  end
+
+  def download(urls = %Stream{}, http_client, path) do
+    Stream.map(urls, &download(&1, http_client, path))
   end
 
   def download(urls, http_client, path) when is_list(urls) do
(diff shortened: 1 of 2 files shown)


Glutexo:
I now think that the `download/3` function that accepts a list of URLs is unnecessary. Traversing an enumerable is a trivial operation and we don’t need to abstract it and separately test it. Here, with the introduction of streams, it shows us that there are more ways to call the `download/3` function with a single URL and until now, it was only limited to lists.

Glutexo:
The aforementioned is addressed in #67.

Glutexo:
Merged current [_elixir_](https://github.com/Glutexo/onigumo) and resolved conflicts.

Glutexo:
[Merged](https://github.com/Glutexo/onigumo/pull/66/commits/9ac77cd353425ad94ab09717c576609f566b558f) current [_elixir_](https://github.com/Glutexo/onigumo/tree/stream) and resolved conflicts. I did not want to duplicate the `download_urls/3` function now when it contains the file name functionality. It stroke me as a bad piece of design. Thus only kept the `load_urls/1` method conversion with no further changes.

nappex (approved):

PR #67, opened 2022-02-12
Remove download_urls/3

Glutexo:
As mentioned in https://github.com/Glutexo/onigumo/pull/66#issuecomment-1037506389, enumerable traversal is a trivial concept. It does not make sense to have a separate function just to do `Enum.map`. It brings confusing function dichotomy and a need for guard clauses. Also it discourages from using other types of enumerables like streams.

For the function to be testable, the `get_env` call needed to be moved to `main/0`, which makes nice symmetry with `http_client`. Because the `input_path` parameter is of the same type and thus indistinguishable from a URL, it only makes sense to rename the function.

### DIFF
--- a/lib/onigumo.ex
+++ b/lib/onigumo.ex
@@ -8,17 +8,16 @@ defmodule Onigumo do
     http_client = Application.get_env(:onigumo, :http_client)
     http_client.start()
 
+    download(http_client, @output_path)
+  end
+
+  def download(http_client, path) do
     Application.get_env(:onigumo, :input_path)
-    |> download_urls(http_client, @output_path)
-  end
-
-  def download_urls(input_path, http_client, output_path) do
-    input_path
     |> load_urls()
-    |> Enum.map(&download_url(&1, http_client, output_path))
+    |> Enum.map(&download(&1, http_client, path))
   end
 
-  def download_url(url, http_client, path) do
+  def download(url, http_client, path) do
     %HTTPoison.Response{
       status_code: 200,
       body: body
--- a/lib/onigumo.ex
+++ b/lib/onigumo.ex
@@ -8,20 +8,17 @@ defmodule Onigumo do
     http_client = Application.get_env(:onigumo, :http_client)
     http_client.start()
 
-    download(http_client, @output_path)
-  end
-
-  def download(http_client, path) do
     Application.get_env(:onigumo, :input_path)
+    |> download_urls(http_client, @output_path)
+  end
+
+  def download_urls(input_path, http_client, output_path) do
+    input_path
     |> load_urls()
-    |> download(http_client, path)
+    |> Enum.map(&download_url(&1, http_client, output_path))
   end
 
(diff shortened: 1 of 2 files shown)


Glutexo:
Extracted the `get_env` move to #68. I’d review and merge that Pull Request first before getting back to this one.

Glutexo:
Reverted the `get_env` move to minimize the amount of changes. Used `Enum.map` in the malfunctioning “download URLs from the input file” test to at least make it still download all the URLs.

Glutexo:
Merged current [_elixir_](https://github.com/Glutexo/onigumo/tree/remove-download-urls) and resolved conflicts.

lib/onigumo.ex
```diff
@@ -13,17 +13,13 @@ defmodule Onigumo do
   end
 
   def download(http_client, dir_path) do
+    file_path = Path.join(dir_path, @output_file_name)
```
Glutexo, line 16: I don’t really like this. But where else should this belong?

test/onigumo_test.exs
```diff
@@ -50,11 +35,12 @@ defmodule OnigumoTest do
...
-    download_result = Onigumo.download(@urls, HTTPoisonMock, tmp_dir)
-    expected_responses = Enum.map(@urls, fn _ -> :ok end)
-    assert(download_result == expected_responses)
-
     output_path = Path.join(tmp_dir, @output_path)
+    Enum.map(@urls, fn url ->
+      download_result = Onigumo.download(url, HTTPoisonMock, output_path)
+      assert(download_result == :ok)
+    end)
+
```
Glutexo, line 43: The test still does not do what it should. Absence of `download` for a list of URLs only exposes this more.

Glutexo:
May become better after #70 is resolved.

test/onigumo_test.exs
Glutexo: Now, after [merging](https://github.com/Glutexo/onigumo/pull/67/commits/278d1e0a7b666c573b0912ab28bcb6c3c22a565c) current [_elixir_](https://github.com/Glutexo/onigumo/tree/elixir) containing #68, the tests works as originally intended without any further changes. 👍🏻

Glutexo:
[Merging](https://github.com/Glutexo/onigumo/pull/67/commits/278d1e0a7b666c573b0912ab28bcb6c3c22a565c) current [_elixir_](https://github.com/Glutexo/onigumo/tree/elixir), resolved conflicts. The Pull Request is now smaller as it does not require changes to the non-working test.

lib/onigumo.ex
nappex:
```suggestion
    output = Path.join(dir_path, @output_file_name)
```

Maybe ?

nappex (approved):

lib/onigumo.ex
Glutexo: This was not about the name, but about the location. But there is no better place for now when the name is not dynamic. It does not make sense to generate the same filename for every URL.

PR #68, opened 2022-02-12
Move get input path to main

Glutexo:
It is not possible to pass a custom input file to the download function and thus the function is not testable. The existing test in fact does not test the file read at all. Fixed by moving the `get_env` call to the main method, which makes a nice symmetry with `http_client`.

Now, the test really tests what it should, the configuration loads are collected to a single place and the functions are even more atomic.

The input path is a string just as a URL and thus these two are indistinguishable. Renamed the function to `download_urls` and `download_url` to avoid this. Still using the ugly guards to distinguish between the URL list and an input file path. Addressing that would be over the scope of this changeset.

Extracted from #67 to make the changeset as small as possible.

### DIFF
--- a/lib/onigumo.ex
+++ b/lib/onigumo.ex
@@ -12,7 +12,7 @@ defmodule Onigumo do
     |> download_urls(http_client, @output_path)
   end
 
-  def download_urls(input_path, http_client, output_path) when is_bitstring(input_path) do
+  def download_urls(input_path, http_client, output_path) when is_binary(input_path) do
     input_path
     |> load_urls()
     |> download_urls(http_client, output_path)
@@ -22,7 +22,7 @@ defmodule Onigumo do
     Enum.map(urls, &download_url(&1, http_client, path))
   end
 
-  def download_url(url, http_client, path) when is_binary(url) do
+  def download_url(url, http_client, path) do
     %HTTPoison.Response{
       status_code: 200,
       body: body
--- a/lib/onigumo.ex
+++ b/lib/onigumo.ex
@@ -10,16 +10,16 @@ defmodule Onigumo do
 
     dir_path = File.cwd!()
     Application.get_env(:onigumo, :input_path)
-    |> download_urls(http_client, dir_path)
+    |> download_urls_from_file(http_client, dir_path)
   end
 
-  def download_urls(input_path, http_client, dir_path) when is_binary(input_path) do
+  def download_urls_from_file(input_path, http_client, dir_path) do
     input_path
     |> load_urls()
     |> download_urls(http_client, dir_path)
   end
 
-  def download_urls(urls, http_client, dir_path) when is_list(urls) do
+  def download_urls(urls, http_client, dir_path) do
     file_path = Path.join(dir_path, @output_file_name)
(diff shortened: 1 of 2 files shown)


lib/onigumo.ex
```diff
@@ -8,20 +8,21 @@ defmodule Onigumo do
...
     http_client.start()
 
-    download(http_client, @output_path)
+    Application.get_env(:onigumo, :input_path)
+    |> download_urls(http_client, @output_path)
   end
 
-  def download(http_client, path) do
-    Application.get_env(:onigumo, :input_path)
+  def download_urls(input_path, http_client, output_path) when is_binary(input_path) do
```
nappex, line 15:
```suggestion
  def download_urls_from_file(input_path, http_client, output_path) when is_binary(input_path) do
```

lib/onigumo.ex
```diff
@@ -8,20 +8,21 @@ defmodule Onigumo do
     http_client = Application.get_env(:onigumo, :http_client)
     http_client.start()
 
-    download(http_client, @output_path)
+    Application.get_env(:onigumo, :input_path)
+    |> download_urls(http_client, @output_path)
```
nappex, line 12:
```suggestion
    |> download_urls_from_file(http_client, @output_path)
```

test/onigumo_test.exs
```diff
@@ -49,7 +49,7 @@ defmodule OnigumoTest do
     File.write!(input_path, content)
 
     output_path = Path.join(tmp_dir, @output_path)
-    result = Onigumo.download(@urls, HTTPoisonMock, output_path)
+    result = Onigumo.download_urls(input_path, HTTPoisonMock, output_path)
```
nappex, line 52:
```suggestion
    result = Onigumo.download_urls_from_file(input_path, HTTPoisonMock, output_path)
```

nappex (commented):
Update the name to more obvious form

Glutexo:
[Merged](https://github.com/Glutexo/onigumo/pull/68/commits/4ece37990203b4640714dda8698d3abca0e28ec1) current [_elixir_](https://github.com/Glutexo/onigumo/tree/elixir) and resolved conflicts.

Glutexo:
[Renamed](https://github.com/Glutexo/onigumo/pull/68/commits/6fc54e1e8b1aa493f2c763dcf02459b9adaadd77) `download_urls/3` that accepts a file to `download_urls_from_file/3`.
 
The two `download_urls` functions do different things. Used a longer name for the one that accepts a file to communicate that it actually does two things – loads a file and download URLs – and that it may be worth further refactoring. This may disappear naturally when leveraging Streams over lists. This is partially addressed by #66.

Thanks @nappex for your review and suggesting the rename!

lib/onigumo.ex
Glutexo: Renamed in https://github.com/Glutexo/onigumo/pull/68/commits/8c7b6716af094a5415c999be9835ad2d72630bcc.

Glutexo:
Please review once more, @nappex now when the conflicts and review suggestions are resolved. Feel free to merge if ✅ .

nappex (approved):

PR #69, opened 2022-02-12
Remove a wrong test

Glutexo:
The “download URLs from the input file” test actually did not test anything else than the “download multiple URLs” test. Although a URLs input file is created, it is not passed to the `download` function and not used at all.

It is currently not possible to test this, because the download function does not accept an input file as an argument. Removing the test until that is fixed.

### DIFF
--- a/lib/onigumo.ex
+++ b/lib/onigumo.ex
@@ -9,21 +9,22 @@ defmodule Onigumo do
     http_client.start()
 
     dir_path = File.cwd!()
-    download(http_client, dir_path)
-  end
-
-  def download(http_client, dir_path) do
     Application.get_env(:onigumo, :input_path)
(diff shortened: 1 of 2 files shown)


Glutexo:
#68 makes the `download` function accept an input file as an argument and thus makes the test work again.

Glutexo:
[Merged](https://github.com/Glutexo/onigumo/pull/69/commits/7f74743e03c9fec8492fa60e57fa32d8b53897a0) current [_elixir_](https://github.com/Glutexo/onigumo/tree/elixir) and resolved conflicts.

Glutexo:
#68 got merged and thus merging in the current [_elixir_](https://github.com/Glutexo/onigumo/tree/elixir) resulted in the test not being necessary to be removed. Closing as this Pull Request became empty.

PR #71, opened 2022-02-13
Convert download_urls_from_file/3 to a Stream

Glutexo:
A Stream allows to process the URL list without collecting responses to a new in-memory list. Removed thus checking for those responses in tests. They are redundant, because `File.write!/3` always returns `:ok`. On failure, it raises an exception.

Complement to #66.


Glutexo:
#66 has been merged. Merged current [_elixir_](https://github.com/Glutexo/onigumo/tree/elixir) into this and resolved conflicts.

nappex (approved):

PR #72, opened 2022-02-13
Split download_url to get and write

### LINKED ISSUE #19: Separate download func to smaller units

Glutexo:
To make the functions as atomic as possible split the `download_url/3` function to smaller chunks. Now, with `Enum.map` used, this is heavily inefficient as it composes a list of all responses in the memory. This can be however resolved by using [Streams](https://hexdocs.pm/elixir/Stream.html). See #66 and #71.

Fixes #19.

### DIFF
--- a/lib/onigumo.ex
+++ b/lib/onigumo.ex
@@ -21,9 +21,13 @@ defmodule Onigumo do
 
   def download_urls(urls, http_client, dir_path) do
     file_path = Path.join(dir_path, @output_file_name)
-    urls
-    |> Enum.map(&get_url(&1, http_client))
-    |> Enum.map(&save_response(&1, file_path))
+    Enum.map(urls, &download_url(&1, http_client, file_path))
+  end
+
+  def download_url(url, http_client, file_path) do
+    url
+    |> get_url(http_client)
+    |> save_response(file_path)
   end
 
   def get_url(url, http_client) do
--- a/lib/onigumo.ex
+++ b/lib/onigumo.ex
@@ -21,16 +21,21 @@ defmodule Onigumo do
 
   def download_urls(urls, http_client, dir_path) do
     file_path = Path.join(dir_path, @output_file_name)
-    Enum.map(urls, &download_url(&1, http_client, file_path))
+    urls
+    |> Enum.map(&get_url(&1, http_client))
+    |> Enum.map(&save_response(&1, file_path))
   end
 
-  def download_url(url, http_client, file_path) do
+  def get_url(url, http_client) do
     %HTTPoison.Response{
       status_code: 200,
       body: body
     } = http_client.get!(url)
+    body
+  end
 
-    File.write!(file_path, body)
+  def save_response(response, file_path) do
+    File.write!(file_path, response)
   end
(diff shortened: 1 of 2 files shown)


Glutexo:
Converted to a draft. The new functions are untested.

Glutexo:
Removed changes from #67.

test/onigumo_test.exs
```diff
@@ -70,6 +70,39 @@ defmodule OnigumoTest do
     assert(loaded_urls == @urls)
   end
 
+  test("get response by HTTP request") do
+    expect(HTTPoisonMock, :get!, &get!/1)
+
+    url = Enum.at(@urls, 0)
+    get_response =  Onigumo.get_url(url, HTTPoisonMock)
```
Glutexo:
```suggestion
    get_response = Onigumo.get_url(url, HTTPoisonMock)
```

test/onigumo_test.exs
```diff
@@ -70,6 +70,39 @@ defmodule OnigumoTest do
...
+    }
+    get_body = Onigumo.get_body(response)
+    expected_body = body(url)
+    assert(get_body == expected_body)
+  end
+
+  @tag :tmp_dir
+  test("write response to file", %{tmp_dir: tmp_dir}) do
+    output_path = Path.join(tmp_dir, @output_path)
+    response = "Response!"
```
Glutexo, line 96: I’d swap these two lines so their order would match the argument order of `write_response/2` below.

test/onigumo_test.exs
```diff
@@ -70,6 +70,36 @@ defmodule OnigumoTest do
...
 
+  test("get response by HTTP request") do
+    expect(HTTPoisonMock, :get!, &get!/1)
+
+    url = Enum.at(@urls, 0)
+    get_response = Onigumo.get_url(url, HTTPoisonMock)
+    expected_response = %HTTPoison.Response{
+      status_code: 200,
+      body: body(url)
+    }
```
Glutexo, line 81: Can we leverage the `get!/1` function down there? The name is then not the best, but it composes the exact response we use here. Just as you do in the “extract body from URL response” test.

lib/onigumo.ex
```diff
@@ -25,12 +25,21 @@ defmodule Onigumo do
...
+    url
+    |> get_url(http_client)
+    |> save_response(file_path)
+  end
+
+  def get_url(url, http_client) do
     %HTTPoison.Response{
       status_code: 200,
       body: body
     } = http_client.get!(url)
```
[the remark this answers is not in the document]
Glutexo: It was. 😃

nappex (approved):

PR #73, opened 2022-02-17
✨ Base64 encode the filename

Glutexo:
A successor to #49.

### DIFF
--- a/lib/onigumo.ex
+++ b/lib/onigumo.ex
@@ -18,8 +18,8 @@ defmodule Onigumo do
   end
 
   def download_url(url, root_path) do
-    filename = create_filename(url)
-    file_path = Path.join(root_path, filename)
+    file_name = create_file_name(url)
+    file_path = Path.join(root_path, file_name)
     url
     |> get_url()
     |> get_body()
@@ -52,7 +52,7 @@ defmodule Onigumo do
     Application.get_env(:onigumo, :http_client)
   end
 
-  def create_filename(url) do
+  def create_file_name(url) do
     Base.url_encode64(url, padding: false)
   end
 end
--- a/lib/onigumo.ex
+++ b/lib/onigumo.ex
@@ -18,7 +18,8 @@ defmodule Onigumo do
   end
 
   def download_url(url, root_path) do
-    file_path = Path.join(root_path, filename(url))
+    filename = create_filename(url)
+    file_path = Path.join(root_path, filename)
     url
     |> get_url()
     |> get_body()
@@ -51,7 +52,7 @@ defmodule Onigumo do
     Application.get_env(:onigumo, :http_client)
   end
 
-  def filename(url) do
+  def create_filename(url) do
     Base.url_encode64(url, padding: false)
   end
 end
--- a/lib/onigumo.ex
+++ b/lib/onigumo.ex
@@ -2,7 +2,6 @@ defmodule Onigumo do
   @moduledoc """
   Web scraper
   """
-  @output_file_name "body.html"
 
   def main() do
     http_client = Application.get_env(:onigumo, :http_client)
@@ -20,8 +19,11 @@ defmodule Onigumo do
   end
 
   def download_urls(urls, http_client, dir_path) do
-    file_path = Path.join(dir_path, @output_file_name)
-    Enum.map(urls, &download_url(&1, http_client, file_path))
+    for url <- urls do
+      file_name = filename(url)
+      file_path = Path.join(dir_path, file_name)
+      download_url(url, http_client, file_path)
+    end
   end
 
   def download_url(url, http_client, file_path) do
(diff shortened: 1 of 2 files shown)


test/onigumo_test.exs
```diff
@@ -3,19 +3,18 @@ defmodule OnigumoTest do
   import Mox
 
   @urls [
-    "http://onigumo.local/hello.html",
-    "http://onigumo.local/bye.html"
+    {"http://onigumo.local/hello.html", "aHR0cDovL29uaWd1bW8ubG9jYWwvaGVsbG8uaHRtbA"},
+    {"http://onigumo.local/bye.html", "aHR0cDovL29uaWd1bW8ubG9jYWwvYnllLmh0bWw"}
```
Glutexo, line 7: This is probably not a good solution. We don’t want to test the encoding algorithm here. We trust the `Base` module and can use it in the test too. It makes the tests more complex as well. Using a Struct would be even better, but that would again increase complexity.

test/onigumo_test.exs
```diff
@@ -47,7 +48,9 @@ defmodule OnigumoTest do
 
     input_path_env = Application.get_env(:onigumo, :input_path)
     input_path_tmp = Path.join(tmp_dir, input_path_env)
-    input_file_content = Enum.map(@urls, &(&1 <> "\n")) |> Enum.join()
+    input_file_content = urls()
+    |> Enum.map(&(&1 <> "\n"))
```
Glutexo, line 52: The capture here is a result of an improper merge. Fixed in #74.

test/onigumo_test.exs
```diff
@@ -3,19 +3,18 @@ defmodule OnigumoTest do
...
   setup(:verify_on_exit!)
 
   @tag :tmp_dir
   test("download a single URL", %{tmp_dir: tmp_dir}) do
     expect(HTTPoisonMock, :get!, &get!/1)
 
-    input_url = Enum.at(@urls, 0)
-    output_path = Path.join(tmp_dir, @output_path)
+    input_url = urls() |> Enum.at(0)
+    output_path = Path.join(tmp_dir, "body.html")
```
Glutexo, line 17: Now, the file name is arbitrary and is used only on this very place. Thus switching to a literal.

test/onigumo_test.exs
Glutexo: #74 is now merged.

lib/onigumo.ex
```diff
@@ -52,4 +50,8 @@ defmodule Onigumo do
   defp http_client() do
     Application.get_env(:onigumo, :http_client)
   end
+
+  def filename(url) do
```
nappex, line 54:
```suggestion
  def encode_url(url) do
```

nappex (commented):
Conflicts resolved, please check it.

Added some proposal improvements...

nappex (approved):

test/onigumo_test.exs
```diff
@@ -108,4 +104,12 @@ defmodule OnigumoTest do
   defp body(url) do
     "Body from: #{url}\n"
   end
+
+  defp assert_downloaded(url, tmp_dir) do
+    filename = Base.url_encode64(url, padding: false)
+    output_path = Path.join(tmp_dir, filename)
```
nappex, line 110:
```suggestion
    output_path = Path.join(tmp_dir, file_name)
```

test/onigumo_test.exs
```diff
@@ -14,39 +14,22 @@ defmodule OnigumoTest do
...
 
     input_url = Enum.at(@urls, 0)
-    output_path = Path.join(tmp_dir, "body.html")
-    download_result = Onigumo.download_url(
-      input_url, HTTPoisonMock, output_path
-    )
+    download_result = Onigumo.download_url(input_url, tmp_dir)
     assert(download_result == :ok)
 
+    output_path = Path.join(tmp_dir, Onigumo.filename(input_url))
```
Glutexo:
```suggestion
    output_filename = Onigumo.filename(input_url)
    output_path = Path.join(tmp_dir, output_filename)
```

test/onigumo_test.exs
```diff
@@ -98,7 +81,8 @@ defmodule OnigumoTest do
   @tag :tmp_dir
   test("write response to file", %{tmp_dir: tmp_dir}) do
     response = "Response!"
-    output_path = Path.join(tmp_dir, @output_path)
+    file_name = "testfile"
```
Glutexo:
```suggestion
    file_name = "body.html"
```

```suggestion
    output_filename = "testfile"
```

lib/onigumo.ex
Glutexo:
```suggestion
  def url_to_filename(url) do
```

```suggestion
  def create_filename(url) do
```

nappex (approved):

Glutexo:
Thanks for merging the conflict and for your suggestions. This was an awesome piece of work!

PR #80, opened 2022-02-20
♻️ Move get input path to load_urls

Glutexo:
There is no practical need to have the URLs input path to be passed to the function as an argument. There is only going to be a single URL input file and the tests too can load the value from the environment.

The directory path still needs to be passed around because the executive code uses the current working directory, whilst the tests use a temporary directory.

### DIFF
--- a/lib/onigumo.ex
+++ b/lib/onigumo.ex
@@ -9,15 +9,13 @@ defmodule Onigumo do
     http_client.start()
 
     dir_path = File.cwd!()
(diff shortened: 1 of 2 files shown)


Glutexo:
Merged in current [_elixir_](https://github.com/Glutexo/onigumo/tree/elixir) and resolved conflicts.

nappex (approved):

Glutexo:
Thx!

PR #82, opened 2022-03-08
♻️ Extract file name creation

Glutexo:
Extract file name creation from a URL to a new function. A follow-up to #73.

### DIFF
--- a/test/onigumo_test.exs
+++ b/test/onigumo_test.exs
@@ -96,6 +96,17 @@ defmodule OnigumoTest do
     assert(read_output == response)
   end
 
+  test("create file name from URL") do
+    input_url = "https://onigumo.local/hello.html"
+    created_file_name = Onigumo.create_file_name(input_url)
+
+    expected_file_name = Base.url_encode64(input_url, padding: false)
+    assert(created_file_name == expected_file_name)
+
+    unexpected_file_name = Base.url_encode64(input_url, padding: true)
+    assert(created_file_name != unexpected_file_name)
+  end
+
   defp prepare_response(url) do
     %HTTPoison.Response{
       status_code: 200,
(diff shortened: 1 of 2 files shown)


test/onigumo_test.exs
```diff
@@ -96,6 +96,17 @@ defmodule OnigumoTest do
     assert(read_output == response)
   end
 
+  test("create file name from URL") do
+    input_url = "https://onigumo.local/hello.html"
```
Glutexo, line 100: The URL is inlined here instead of using `@urls`, because we need a URL that is encoded with padding. For example `bye.html` would be encoded into a string without padding and thus it would not be check that `padding: false` is used.

test/onigumo_test.exs
```diff
@@ -96,6 +96,17 @@ defmodule OnigumoTest do
...
+    created_file_name = Onigumo.create_file_name(input_url)
+
+    expected_file_name = Base.url_encode64(input_url, padding: false)
+    assert(created_file_name == expected_file_name)
+
+    unexpected_file_name = Base.url_encode64(input_url, padding: true)
+    assert(created_file_name != unexpected_file_name)
```
Glutexo, line 107: This fails if `input_url` is changed to a value that does not yield encoded string with a padding.

test/onigumo_test.exs
nappex: OK, that's make sense when we want to ensure some behavior.
I have idea to take first url from `@urls`
```suggestion
    input_url = Enum.at(@urls, 0)
```

But your approach seems to me safer, when urls will changed in furutre and forgot to this test.

test/onigumo_test.exs
nappex: Nice, very clever fuse. :+1:

nappex (approved):

PR #83, opened 2022-03-08
♻️ Extract download assertion

Glutexo:
To have the download test shorter and not to use an overly long anonymous function, extracted the faily complex verification of the downloaded URL to a function. This function can be easily put into `Enum.map`.

Extracted from #73.

### DIFF
--- a/test/onigumo_test.exs
+++ b/test/onigumo_test.exs
@@ -35,13 +35,7 @@ defmodule OnigumoTest do
 
     Onigumo.download_urls_from_file(tmp_dir) |> Stream.run()
 
-    Enum.map(@urls, fn url ->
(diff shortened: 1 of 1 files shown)


test/onigumo_test.exs
```diff
@@ -111,4 +105,12 @@ defmodule OnigumoTest do
   defp body(url) do
     "Body from: #{url}\n"
   end
+
+  defp assert_downloaded(url, tmp_dir) do
+    file_name = Base.url_encode64(url, padding: false)
```
Glutexo, line 110: This is going to be replaced with `Onigumo.create_filename` once #82 is merged. This function has its own test and as such it is trusted.

nappex (approved):

PR #87, opened 2022-03-27
Add parser of html links

nappex:
Parser which is able to convert string document to html structure, find the `a` tag a then get the href value.


nappex:
Viva la SIGIL , I dont know why

Glutexo:
😮 This is going to be the foundation of the PyCZ plugin. 👍🏻 I don’t know where this should belong yet, but a plain module may be enough for now.

Glutexo:
Or maybe a part of a generic link finding plugin. 🤔  But, it may be too challenging to develop a plugin architecture. I’d put that to the Spider and think about extracting it to a plugin later.

Glutexo (changes requested):
After thinking more, I concluded that providing a set of the most commonly used scraping tools would make the project much more helpful. The goal of Onigumo is to automate the common ground for scraping and to provide a toolchain nicely fits.

For now:
This functionality does not belong to Onigumo Parser. Let’s move it to some Plugins or Tools namespace and name it more precisely, e.g., LinkExtractor, LinkFinder, Links, or similar.
Rebase on the current elixir to clean the diff of the Hash-related changes.
After that, I think this will be up to a proper review. Thanks for the excellent work!

Glutexo:
I [merged](https://github.com/Glutexo/onigumo/pull/87/commits/17612b205968e618eaaa56b5c6b53f13e1f38b9a) in the current master to minimize the number of conflicts. The pull request is now clean of unrelated changes.

Glutexo:
You expressed uncertainty on what to do with this pull request. As mentioned in my [review](https://github.com/Glutexo/onigumo/pull/87#issuecomment-1190443917), I think the link finder belongs instead of the Parser to a standalone module in a toolbox/plugin namespace.

I can’t think now of an ideal architecture. Thus I won’t object to anything that isn’t wrong. The Parser’s job is to check for the downloaded raw data and pass it to the Spider’s concrete parser, not to do the parsing itself.

Is it clear now? Don’t hesitate to raise any questions.

The changeset is clear after the mer

lib/parser.ex
```diff
@@ -0,0 +1,7 @@
+defmodule Parser do
+  def html_links(document) do
+    Floki.parse_document!(document)
+    |> Floki.find("a")
+    |> Floki.attribute("href")
```
Glutexo, line 5: What happens if the anchor does not have a `href` attribute?

Glutexo, line 5: This is actually interestingThe way how Floki operates on collections is interesting. I like it.

mix.exs
```diff
@@ -25,6 +25,7 @@ defmodule Onigumo.MixProject do
       # {:dep_from_hexpm, "~> 0.3.0"},
       # {:dep_from_git, git: "https://github.com/elixir-lang/my_dep.git", tag: "0.1.0"},
       {:httpoison, "~> 1.8"},
+      {:floki, "~> 0.32.0"},
```
Glutexo, line 28: I’d consider creating a section with a heading describing that the following dependencies are not for the core but the tools.

Glutexo, line 28: I’d consider creating a section with a heading describing that the following dependencies are not for the core but the tools.

test/parse_test.exs
```diff
@@ -0,0 +1,25 @@
...
+      <html>
+      <body>
+        <section id="content">
+          <p class="headline">Floki</p>
+          <a href="http://onigumo.local/hello.html">Hello</a>
+          <a href="http://onigumo.local/bye.html">Bye</a>
+          <span data-model="user">onigumo</span>
+        </section>
+      </body>
+      </html>")
```
Glutexo, line 20: Why are URLs in a module constant but the document in a function variable? I can’t tell which is better, but I feel like having it consistent would be better. I am personally more inclined to module constants. We already did so in the Hash [test](https://github.com/nappex/onigumo/blob/links-parser/test/hash_test.exs).

Glutexo, line 20:
```suggestion
```

test/parse_test.exs
```diff
@@ -0,0 +1,25 @@
...
+
+    result = Parser.html_links(document)
```
Glutexo, line 22:
```suggestion
    result = Parser.html_links(@document)
```

test/parse_test.exs
```diff
@@ -0,0 +1,25 @@
+defmodule ParseTest do
+  use ExUnit.Case
+
+  @urls [
```
Glutexo, line 4:
```suggestion
  @document = ~s("<!doctype html>
    <html>
    <body>
      <section id="content">
        <p class="headline">Floki</p>
        <a href="http://onigumo.local/hello.html">Hello</a>
        <a href="http://onigumo.local/bye.html">Bye</a>
        <span data-model="user">onigumo</span>
      </section>
    </body>
    </html>")
  @urls [
```

test/parse_test.exs
```diff
@@ -0,0 +1,25 @@
...
+    "http://onigumo.local/hello.html",
+    "http://onigumo.local/bye.html"
+  ]
+
+  test("Parsing values of href attributes in html links") do
+    document = ~s("<!doctype html>
```
Glutexo, line 10: Does the `~s` sigil trim the whitespace from the beginning on the following lines? Does the string contain `<!doctype html>\n<html>…` or `<!doctype html>\n      <html>…`? The former is correct; the latter is not.

Glutexo (changes requested):
I am only requesting changes because of the module name. As mentioned before, this functionality is not Onigumo’s Parser. All other comments are only suggestions or questions, and I don’t need to block the pull request because of them. We can improve on them in follow-up ones.

mix.exs
nappex: I added

lib/parser.ex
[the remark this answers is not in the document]
nappex: Thanks for the point. I've tested and Floki doesnt invoke an error if the `a` tag does not contain a `href`.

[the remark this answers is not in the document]
Glutexo: 🤨

mix.exs
```diff
@@ -25,7 +25,10 @@ defmodule Onigumo.MixProject do
       # {:dep_from_hexpm, "~> 0.3.0"},
       # {:dep_from_git, git: "https://github.com/elixir-lang/my_dep.git", tag: "0.1.0"},
       {:httpoison, "~> 1.8"},
-      {:mox, "~> 1.0", only: :test}
+      {:mox, "~> 1.0", only: :test},
+
+      # Toolbox dependencies
+      {:floki, "~> 0.32.0"}
```
Glutexo, line 31: ❤️

Glutexo (changes requested):
I have only one real objection, marked with ⚠️. Everything else is only light suggestions. Great job! 💪🏻

test/parse_test.exs
Glutexo: The [s sigil](https://hexdocs.pm/elixir/1.14/Kernel.html#sigil_s/2) does not trim whitespace from the beginning of the line. Although the indentation looks nice in the code, it would incorrectly propagate to the inserted HTML. HTML is insensitive to whitespace, so the indentation does not cause any harm, but it still is an unwanted interference.

mix.exs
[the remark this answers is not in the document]
Glutexo: Oh, you mentioned that in https://github.com/Glutexo/onigumo/pull/161#issuecomment-1383179674.

Glutexo:
I also manually tested the new [_find_links_](https://github.com/nappex/onigumo/blob/af9da826466bb47486189b5330368350410c41a4/lib/spider_html.ex#L2) function, and it works well:

```shell
$ iex -S mix
```
```elixir
iex(1)> Spider.HTML.find_links(~s(<a href="http://www.seznam.cz/"></a><a class="link" href="http://www.mapy.cz/"></a><a id="nothing"></a>))        
["http://www.seznam.cz/", "http://www.mapy.cz/"]
```

test/spider_html_test.exs
```diff
@@ -0,0 +1,27 @@
...
+            <a href="http://onigumo.local/hello.html">Hello</a>
+            <a href="http://onigumo.local/bye.html">Bye</a>
+            <a id="b"></a>
+            <span data-model="user">onigumo</span>
+        </section>
+    </body>
+</html>)
+
+  describe("Spider.HTML.find_links/1") do
+    test("Parsing values of href attributes in html links") do
```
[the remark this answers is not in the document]
nappex:
```suggestion
    test("find links in href attributes of 'a' tags") do
```

test/spider_html_test.exs
```diff
@@ -0,0 +1,27 @@
...
+  ]
+  @html ~s(<!doctype html>
+<html>
+    <body>
+        <section id="content">
+            <p class="headline">Floki</p>
+            <a href="http://onigumo.local/hello.html">Hello</a>
+            <a href="http://onigumo.local/bye.html">Bye</a>
+            <a id="b"></a>
+            <span data-model="user">onigumo</span>
```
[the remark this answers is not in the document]
Glutexo: We could utilize a test verifying that a non “a” tag with a “href” attribute does not appear in the result. Such tag is not valid HTML, it can still appear in the real world, though. We can discuss, whether finding hrefs on other tags should be in the result, but the current implementation doesn’t do so, so it needs have a test. See #170.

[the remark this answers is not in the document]
Glutexo: My previous claim was false: there are more elements that can have the _href_ attribute: _area_, _link_ etc.

nappex:
Elixir library `Floki` uses `parse_document` or `parse_fragment`. By default there is no difference between them because `parse_fragment` is just wrapper of `parse_document`. But this behaviour is just for default `Floki.HTMLParser`, you could use your own HTMLParser which defined different behave of these functions. 

From [documentation of Floki](https://hexdocs.pm/floki/Floki.html#parse_document/2)
`parse_document(document, opts \\ [])` the opts is for specification of any `HTMLParser`

See:
```elixir
iex>Floki.parse_document("<html><head></head><body>hello</body></html>")
iex>{:ok, [{"html", [], [{"head", [], []}, {"body", [], ["hello"]}]}]}

iex>Floki.parse_document("<html><head></head><body>hello</body></html>", html_parser: Floki.HTMLParser.Mochiweb)
iex>{:ok, [{"html", [], [{"head", [], []}, {"body", [], ["hello"]}]}]}
```

test/spider_html_test.exs
```diff
@@ -0,0 +1,30 @@
...
+  use ExUnit.Case
+
+  @urls [
+    "http://onigumo.local/hello.html",
+    "http://onigumo.local/bye.html"
+  ]
+  @html ~s(<!doctype html>
+<html>
+    <head>
+        <link href="/media/examples/link-element-example.css" rel="stylesheet">
```
Glutexo, line 11: ❤️

Glutexo (approved):
LGTM

PR #91, opened 2022-07-05
🌐 Translate About and Architecture

Glutexo:


nappex:
Just think, does it make you sense to add info about which protocols we're supporting somewhere in README info? I think that support of other protocols instead just HTTP or HTTPS does not make sense but I am not sure....

Glutexo:
> Just think, does it make you sense to add info about which protocols we're supporting somewhere in README info? I think that support of other protocols instead just HTTP or HTTPS does not make sense but I am not sure....

It may make sense in the Downloader section or the general overview. But it feels similar to the JSON dispute earlier: it is too implementation-specific. Also, although we didn’t think about supporting anything other than HTTP, it makes sense to scrape an FTP drive, a Git repository, an IMAP mailbox, or a WebDAV storage.

We don’t support anything other than HTTP for now, but even that is still very rudimentary. It would be bold even to say we support HTTP.

Glutexo:
I started to dislike the changes I proposed. The translation was not accurate and even incorrect. Also, the Pull Request is too big, amending two sections simultaneously. I’ll replace it with better ones.

Thank you for the review!

nappex (approved):

nappex:
Ok, unfortunately my english skills are not up to the level which I could assess quality of your translation. I've just known that it made sense and I understood the content from it.

Glutexo:
Superseded by #104.

PR #93, opened 2022-07-06
♻️ Shorten “URL address“ to just “address”

Glutexo:
I find the “URL address” expression too wordy. I picked “address,” but “URL” would be OK too. In English, I’d instead choose the “URL” version.

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,24 +27,24 @@ flowchart LR
 
 ### Operator ###
 
-Určuje URL adresy ke stažení pro _downloader_. Za přidávání URL adres ke zpracování je zodpovědný pavouk. Nové URL adresy pavouk získává z naparsované podoby dat, kterou vytváří _parser_.
+Určuje adresy ke stažení pro _downloader_. Za přidávání adres ke zpracování je zodpovědný pavouk. Nové adresy pavouk získává z naparsované podoby dat, kterou vytváří _parser_.
 
 Činnost _operatoru_ se skládá z:
 
 1. inicializace pavouka,
-2. kontroly existence seznamu nových URL adres,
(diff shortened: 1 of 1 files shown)


Glutexo:
I gave the text for a review and got the advice that using “URL” instead of just “address” would sound better, even in Czech. After all, this is a technical text. Still, using the complete form in the first appearance may look good.

I pushed a new version, so you can assess how it looks. I kept the full “URL address” form only in the first appearance in each section and just “URL” in all subsequent ones.

README.md
```diff
@@ -27,24 +27,24 @@ flowchart LR
...
-Určuje URL adresy ke stažení pro _downloader_. Za přidávání URL adres ke zpracování je zodpovědný pavouk. Nové URL adresy pavouk získává z naparsované podoby dat, kterou vytváří _parser_.
+Určuje URL adresy ke stažení pro _downloader_. Za přidávání URL ke zpracování je zodpovědný pavouk. Nové URL pavouk získává z naparsované podoby dat, kterou vytváří _parser_.
 
 Činnost _operatoru_ se skládá z:
 
 1. inicializace pavouka,
-2. kontroly existence seznamu nových URL adres,
-3. načítání nových URL adres ze strukturovaných dat,
-4. zařazování těchto URL adres do fronty pro _downloader_.
+2. kontroly existence seznamu nových URB,
```
nappex, line 35:
```suggestion
2. kontroly existence seznamu nových URL,
```

nappex (changes requested):

nappex:
Ok, it sounds good I agree.

README.md
Glutexo: Nice catch!

PR #104, opened 2022-07-13
🌐 Translate About

Glutexo:
Partial replacement for #91.


Glutexo:
I want to make the English version as close to the original text and not mix amendments with translation.

nappex (approved):

PR #109, opened 2022-07-15
Remove redundant Mermaid whitespace

Glutexo:
The aligning whitespace is pretty but hardly sustainable.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,11 +16,11 @@ The flowchart below illustrates the cooperation of those parts:
 
 ```mermaid
 flowchart LR
-    START             -->           operator(OPERATOR)
-    operator    -- <hash>.urls --->    downloader(DOWNLOADER)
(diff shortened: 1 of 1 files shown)


nappex (commented):
it was made for better readability, do you think it has not a worth?

Glutexo:
It certainly had with the very first version of the flowchart. Now, if you want to amend it, you need to reformat whitespace on all the lines, which seems rather impractical. Just look how broken it was before this Pull Request. The position of the top and bottom arrows is off, and so is the alignment of the right side.

nappex (dismissed):
alright, flowchart is not so complicated... to stick on the readability.

Glutexo:
Instead of collapsing the whitespace, I cleaned up the aligned version. Although it looks nice, I am afraid the maintainability won’t be high. But we can revisit the formatting once it becomes a real problem. Thank you for your input!

nappex (approved):

PR #112, opened 2022-07-18
🌐 Translate Operator

Glutexo:
Translate the Operator subsection of the Architecture section in the README. The translation is literal; not paying attention to the style.

I decided to go for a literal translation, even though it sounds awful, to minimize the number of changes. I plan to clean up the style in follow-up pull requests.

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,14 +25,14 @@ flowchart LR
 
 ### Operator ###
 
-Determines URL addresses for download for the Downloader. A Spider is responsible for adding the URLs. The Spider gets the new URLs from the parsed form of the data, which the Parser creates.
+The Operator determines URL addresses for the Downloader. A Spider is responsible for adding the URLs, which it gets from the parsed form of the data provided by the Parser.
 
-The Operator’s job is composed of:
+The Operator’s job to:
 
-1. initialization of a Spider,
+1. initialize a Spider,
 2. check for the existence of a new URL list,
-3. extraction of new URLs from structured data,
-4. insertion of those URLs to the Downloader queue.
+3. extract new URLs from structured data,
+4. insert those URLs onto the Downloader queue.
 
 ### Downloader ###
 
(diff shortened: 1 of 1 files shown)


Glutexo:
I changed the wording so that it does not change the original meaning. The sentences do not need to match 1:1 to the Czech version. I paid attention to keeping the factual contents, not slipping into clarifying them to express my ideas better.

nappex:
your text make sense to me. True is that every PR play telephone..

README.md
```diff
@@ -25,14 +25,14 @@ flowchart LR
...
 
-Činnost _operatoru_ se skládá z:
+The Operator’s job is to:
 
-1. inicializace pavouka,
-2. kontroly existence seznamu nových URL,
-3. načítání nových URL ze strukturovaných dat,
-4. zařazování těchto URL do fronty pro _downloader_.
+1. initialize a Spider,
+2. check for the existence of a new URL list,
```
nappex, line 33:
```suggestion
2. check for the existence of a new URLs,
```

nappex (approved):

README.md
Glutexo: The original version mentions URL lists (“seznam nových URL“), as in files with the .urls extension. That’s why I kept it in the translation.
The suggestion is grammatically incorrect. A plural “URLs” does not conform to the usage of the indefinite article “a.”

```suggestion
2. check for the existence of new URLs,
```

Why the change from the “URL lists” to “URLs”? I don’t want to merge before we resolve this dispute.

nappex: to not be so much specific but yes if we translate czech version then it should be keep it, deletion should be done in czech if so.

Glutexo: I moreover think that we should delete the whole bullet point. **The Downloader** checks the existence of the files with URLs, also called URL lists.


“
1. Initialize a spider,
2. extract new URLs from structured data
”
looks more precise to me.

PR #122, opened 2022-07-26
📝 Swap the word parsed for structured

Glutexo:
Parsing is the process. The original raw data becomes “parsed,” but the result is a “structure.”

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,7 +25,7 @@ flowchart LR
 
 ### Operator ###
 
-The Operator determines URL addresses for the Downloader. A Spider is responsible for adding the URLs, which it gets from the parsed form of the data provided by the Parser.
+The Operator determines URL addresses for the Downloader. A Spider is responsible for adding the URLs, which it gets from the structured form of the data provided by the Parser.
 
 The Operator’s job is to:
 


nappex:
I thought that if you make from unstructured data to structure data the process is calling parsing so it could be usable even both of terms parsed or structure

Glutexo:
“Parsing” is the process of finding the information in the raw input. I see how the output can fit the term “parsed data;” the result is unavoidably structured, be it only in memory. I, however, feel that there is still a slight difference.

For example, an HTML input like this:

```html
<ul id="animals">
  <li>lizard</li>
  <li>snake</li>
  <li>dragon</li>
</ul>
```

Can go through a generic parser similar to that one of yours (#87), yielding an object like this:

```python
>>> parsed_html
{
  "lists": {
    "animals": [
      "lizard",
      "snake",
      "dragon"
    ]
  }
}
```

I’d call that “parsed data” in the sense of “the result of parsing.” But I imagined the parser to store not the immediate result of generic object finding but rather a semantic structure. E.g.:

```json
{
  "animals": [
    "lizard",
    "snake",
    "dragon",
  ]
}
```

The difference is not just the missing “lists” key but the semantics of the data. The former is just a structured version of the original text page; the latter is a set of values for which we are looking and that we will use for further downloading or materialization.

Does the explanation make sense? I can give out a better example if it’s still hazy.

nappex:
maybe, parsed data are strict same as source but written with diff syntax and structured data are diff syntax and core of information is same but not necessary with all meta data?

Glutexo:
That’s a good description! The parsed data is what you get, the structured data is what you need.

nappex (approved):

PR #129, opened 2022-07-31
📝 Add Spider tasks to the flowchart

### LINKED ISSUE #115: 📝 Extend the flowchart with Spider tasks
The flowchart lacks details about how the data flows from Onigumo to a Spider and back. I don’t want the flowchart to become overly complicated, but the distinction between the framework and its apps is crucial, in my opinion.

My idea is something like

> Start → Onigumo Operator → Spider Operator → Downloader → Onigumo Parser → Spider Parser → back to Onigumo Operator → Spider Materialization

The main point here is that Onigumo only determines what to process, and the Spiders do the actual work. Spider Operator receives a list of parsed entities and tells what to download. Spider Parser gets a list of unparsed entities and provides structured data.

There is no parsed data at the beginning, which marks a starting state. When there is nothing more to download, it is time for materialization.

The flowchart may be a bit more colorful to distinguish between the core and the userspace.

Glutexo:
The flowchart was very simple, only mentioning the core parts of the workflow. The distinction between the framework and the userspace needs the diagram to be more complex. I resolved that by adding a subgraph showing the parts that require programming.

Fixes #115.

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,11 +17,9 @@ The flowchart below illustrates the flow of data between those parts:
 ```mermaid
 flowchart LR
     START              -->         operator(OPERATOR)
-
     operator   -- <hash>.urls ---> downloader(DOWNLOADER)
     downloader -- <hash>.raw  ---> parser(PARSER)
     parser     -- <hash>.json ---> operator
-
     operator           -->         materialization(MATERIALIZATION)
 
     subgraph Spider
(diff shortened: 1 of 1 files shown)


Glutexo:
Mermaid is a beast. All my attempts to create more subgraphs or integrate the Spider tasks into the core cycle resulted in a convoluted spaghetti of connecting lines.

nappex (approved):

PR #130, opened 2022-07-31
📝 Make the Start box rounded

Glutexo:
The Start is just a virtual step that is neither part of Onigumo nor a Spider. It is a mere entry-point to the Operator that behaves the same way as during runtime.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,11 +16,11 @@ The flowchart below illustrates the flow of data between those parts:
 
 ```mermaid
 flowchart LR
-    START              -->         operator(OPERATOR)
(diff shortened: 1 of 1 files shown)


nappex:
why do you use this kond of syntax?

Glutexo:
The flowchart syntax documentation says [so](https://mermaid-js.github.io/mermaid/#/flowchart?id=a-stadium-shaped-node). Is there a better way do make the rounded shape?

Or are you asking why the rounded shape? Just to differentiate from the default, which is more square.

nappex (approved):

PR #133, opened 2022-08-07
Make the start bubble round again

Glutexo:
I accidentally made the start point rectangular. It was part of my last patch that changed the component boxes’ shape. I am making the shape of this virtual item round again.


Glutexo:
I am merging this forcefully. @nappex already approved the change in its previous pull request. I removed it only by accident.

PR #134, opened 2022-08-07
♻️ Rename Onigumo to Onigumo.Downloader

### LINKED ISSUE #118: ♻️ Rename Onigumo module to Downloader
The current Onigumo module does the downloading part of the workflow. Other components like the Parser and the Operator become distinct modules. The rename would improve semantics and prevent confusion when more modules appear in the codebase.

Glutexo:
The current Onigumo module does the downloading part of the workflow. Other components like the Parser and the Operator become distinct modules. The rename improves semantics and prevents confusion when more modules appear in the codebase.

Fixes #118.

### DIFF
--- a/lib/onigumo.ex
+++ b/lib/onigumo.ex
@@ -1,4 +1,4 @@
-defmodule Onigumo do
+defmodule Onigumo.Downloader do
   @moduledoc """
   Web scraper
   """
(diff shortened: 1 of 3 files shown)


Glutexo:
@nappex The word “semantics” refers here to the fact that the module’s job is to download. Thus, Downloader is the most appropriate name. Onigumo, on the other hand, is the name of the whole project and in the current context, using it as a module name is vague and potentially misleading.

Feel free to ask for clarification if this description does not fit.

nappex (approved):

PR #136, opened 2022-08-14
♻️ Add describe to tests

Glutexo:
ExUnit’s [_describe_](https://hexdocs.pm/ex_unit/1.12/ExUnit.Case.html#describe/2) function allows grouping or describing tests. I used it to display what executive function the test verifies. The [example](https://hexdocs.pm/ex_unit/1.12/ExUnit.Case.html#describe/2-examples) in the documentation shows this approach.

It also allows testing a specific function by calling:

```shell
$ mix test --only describe:Onigumo.download_url/2
```


Glutexo:
I resolved the conflicts. They appeared because of the rename of the Onigumo module to Onigumo.Downloader. To make the function identifiers match, I added the “Downloader” part to the _describe_ strings.

nappex:
I had to make merge master again locally to solve problem with the README file diff....

Unfortunately I dont know why....

nappex:
When you want to see changes and ignore changes in whitespaces you can follow this article.

If you want to achieve same result in CLI option `-w` is your friend, complete command will be `git diff -w master`

nappex:
I've tested the case in first comment and no test was choosen.

```
$ mix test --only describe:Onigumo.download_url/2
Excluding tags: [:test]
Including tags: [describe: "Onigumo.download_url/2"]



Finished in 0.03 seconds (0.00s async, 0.03s sync)
12 tests, 0 failures, 12 excluded

Randomized with seed 105190
The --only option was given to "mix test" but no test was executed
```

nappex:
We could explore diff between `describe` and `describetag`.

nappex:
I've understand it well, describe just add additional info to test description and we can include/exclude by describes certain tests.

nappex (approved):
Thanks, I think it improved slightly our code and we learn something new :partying_face:

Glutexo:
Thank you for the research on the diffing! I will experiment with the switches a bit. Great work! I am curious about what happened to the Git branch in your local copy. We can dig it out of the reflog if you are interested too. Feel free to reach out to me.

To clarify the goal of this pull request: the describe clause not only adds additional information to the tests but also groups them. That makes it possible to run tests for a given function simply by passing its name, even if there is more than one.

In our case, the parametrization in #142 will remove the need for the grouping. The advantage of describing the tests by the name of the function they target remains.

Please ask if you have any questions.

PR #138, opened 2022-08-14
♻️ Make tests order match the tested module

Glutexo:
For better discoverability of the tests, I changed the test order to match the order of the functions in the executive module.

During the review, please verify the following:

- [x] That the new test order matches the order of the functions in the tested module;
- [x] That the moved tests are identical, without any single change.


Glutexo:
I resolved the conflicts caused by the renaming of the Onigumo module.

test/onigumo_test.exs
```diff
@@ -39,27 +25,18 @@ defmodule OnigumoTest do
   end
 
   @tag :tmp_dir
-  test("load a single URL from a file", %{tmp_dir: tmp_dir}) do
```
nappex, line 42: so we dont test a URL, because it is enough to test multiple URLs is that right?

Glutexo: You are right, but it was my mistake to include the change in this pull request. I mistakenly pasted modified tests from #142. My intention was only to change the order of the tests, not to change them. I reverted the tests to their shape from “master”.

Nice catch, thank you!

Glutexo:
I [merged](https://github.com/Glutexo/onigumo/pull/138/commits/d1fc80c1fb415dc2ca369baaf5f1044b12626959) the current _master_ into this branch and resolved conflicts. Also, I removed a forgotten constant (@slices) from the previously [mentioned](https://github.com/Glutexo/onigumo/pull/138#discussion_r957709249) incorrect merge. The constant raised warnings in the PR check because it had no usage.

Glutexo:
I added my motivation and review criteria to the PR description.

nappex (approved):

PR #139, opened 2022-08-16
✅ Test Onigumo.Downloader.main

### LINKED ISSUE #56: ✅ Test main
This needs good mocks, probably also creating a behavior. Maybe even splitting the Onigumo module.

Glutexo:
The application

- starts the HTTPClient,
- downloads the URLs.

The new test is very similar to [_download_urls_from_file_](https://github.com/Glutexo/onigumo/blob/5d0a6a4f81ab69ccdc2c7df0d74ce89a85d36e99/test/onigumo_test.exs#L43). In this case, the Stream runs automatically, and the test only verifies the result. The assertion of the _Stream.run_ return value is redundant, but in another test case, we check that _File.write!_ returns _:ok_, although there is no other possible return value.

The assertion of _HTTPoison.start_ is mandatory: Mox requires all mock calls to have an _expect_ clause. There is no need to test that separately then.

Fixes #56.


test/onigumo_test.exs
```diff
@@ -9,6 +9,22 @@ defmodule OnigumoTest do
 
   setup(:verify_on_exit!)
 
+  @tag :tmp_dir
+  test("run the download", %{tmp_dir: tmp_dir}) do
+    expect(HTTPoisonMock, :start, fn -> nil end)
```
Glutexo, line 14: This assertion is mandatory. Mox requires all calls to have an _expect_ clause.

test/onigumo_test.exs
```diff
@@ -9,6 +9,22 @@ defmodule OnigumoTest do
...
+    expect(HTTPoisonMock, :get!, length(@urls), &prepare_response/1)
+
+    input_path_env = Application.get_env(:onigumo, :input_path)
+    input_path_tmp = Path.join(tmp_dir, input_path_env)
+    input_file_content = prepare_input(@urls)
+    File.write!(input_path_tmp, input_file_content)
+
+    stream_result = Onigumo.Downloader.main(tmp_dir)
+    assert(stream_result == :ok)
```
Glutexo, line 23: This assertion is redundant and lies here for consistency with [a result](https://github.com/Glutexo/onigumo/blob/5d0a6a4f81ab69ccdc2c7df0d74ce89a85d36e99/test/onigumo_test.exs#L34) of _File.write!_ in another test.

test/onigumo_test.exs
```diff
@@ -9,6 +9,22 @@ defmodule OnigumoTest do
 
   setup(:verify_on_exit!)
 
+  @tag :tmp_dir
+  test("run the download", %{tmp_dir: tmp_dir}) do
```
Glutexo, line 13: I could not find a more suitable description for the test. Do you have any ideas?

test/onigumo_test.exs
nappex: Line 14 is necessary I've tested...

test/onigumo_test.exs
nappex: I am not sure what do you want to say....
I suggest the :ok is return value from `Stream.run()` when I've read the docs - https://hexdocs.pm/elixir/Stream.html#run/1

nappex:
- [ ] Solve return value `:ok`
- [ ] Write better description text for test

test/onigumo_test.exs
nappex: my suggestions to

`Onigumo.Downloader.main`:
- start
- init
- trigger

"run the download"
- activate downloading
- start download process
- integration test for downloader
- check download process

nappex: We discussed personally, there is a option to split func `main` to `init` and `run`, where `start` should be located in `init` and rest in `main`....

nappex: So far we keep the name `main`

Glutexo: We kept `main` because the Downloader is essentialy a standalone application and the _main_ function is its entry points. The caller does not care about what the module does. `main` is a traditional name from the ages of C and Java.

Glutexo: The new name of the tests is „run Downloader,“ because what we test here is that the Downloader does its job. The fact that it loadts the URLs from a file and then downloads them is a mere technical detail.

nappex (approved):

Glutexo:
Your changes are great, @nappex! Feel free to ✅ .

PR #143, opened 2022-08-19
♻️ Move for loops outside the tests

Glutexo:
A loop inside the tests makes all the iterations run in the same test case. That means that if one of them fails, the subsequent ones don’t even run. They also share the environment, which causes no problem here, but as a practice, it can interfere later, e. g. with tests using temporary directories.

Related to #142 and #59.

PR #142, opened 2022-08-17
♻️ Parametrize the load URL test

Glutexo:
The “load a single URL” and “load multiple URLs” tests were almost identical, only dealing with a different set of URLs. Parametrization reduced repetition.

To match the style in [Hash tests](https://github.com/Glutexo/onigumo/blob/6f48cd5b48b49c8cdc3756f4813faa229f34cd78/test/hash_test.exs), the _for_ clause is inside the test block.

Fixes #59.

PR #144, opened 2022-08-19
♻️ Move the loop outside the test

Glutexo:
When outside the test function, the loop creates a test case for every iteration. Like that, if the test fails, other parametrized variations still run. Every one of them gets its descriptive line in the trace.

This pull request is a follow-up to #142, doing the same as #143 with the new test with a loop.

Interestingly, not all values can be simply [unquoted](https://hexdocs.pm/elixir/Kernel.SpecialForms.html#unquote/1). For technical reasons, [Macro.escape/1](https://hexdocs.pm/elixir/Macro.html#escape/2) must wrap the [unquoted](https://hexdocs.pm/elixir/Kernel.SpecialForms.html#unquote/1) value to unquote a Range.


test/onigumo_test.exs
```diff
@@ -39,14 +39,13 @@ defmodule OnigumoTest do
...
-  test("load URLs from a file", %{tmp_dir: tmp_dir}) do
-    input_path_env = Application.get_env(:onigumo, :input_path)
-    input_path_tmp = Path.join(tmp_dir, input_path_env)
-
-    for slice <- @slices do
-      input_urls = Enum.slice(@urls, slice)
+  for slice <- @slices do
+    @tag :tmp_dir
+    test("load URLs #{inspect(slice)} from a file", %{tmp_dir: tmp_dir}) do
+      input_urls = Enum.slice(@urls, unquote(Macro.escape(slice)))
```
Glutexo, line 45: I resorted to putting the whole complex expression `unquote(Macro.escape(slice))` inline instead of assigning its result to an intermediary variable. I understood it as a wrapper around the name to use the variable and tried to hide the logic behind it. If the variable appeared in multiple places, I might have chosen differently.

```suggestion
      slice_param = unquote(Macro.escape(slice))
      input_urls = Enum.slice(@urls, slice_param)
```

nappex (approved):

test/onigumo_downloader_test.exs
```diff
@@ -94,14 +94,13 @@ defmodule OnigumoDownloaderTest do
...
-    test("load URLs from a file", %{tmp_dir: tmp_dir}) do
-      input_path_env = Application.get_env(:onigumo, :input_path)
-      input_path_tmp = Path.join(tmp_dir, input_path_env)
-
-      for slice <- @slices do
-        input_urls = Enum.slice(@urls, slice)
+    for slice <- @slices do
+      @tag :tmp_dir
+      test("load URLs #{inspect(slice)} from a file", %{tmp_dir: tmp_dir}) do
+        input_urls = Enum.slice(@urls, unquote(Macro.escape(slice)))
```
nappex, line 100: How works `unquote` in combination with `Macro.escape`:

`unquote` is macro, which wrap `Macro.escape` (escape is function of module Macro, module which contain helpful functions for working with macros).

`unquote` put content of variable slice to `Macro.escape` function, we embed range 0..1 or 0..-1 instead of name of variable which will not be defined in frame of test function after compilation process. Finally `Macro.escape` creates AST for complication data structure as `range`.

`unquote` can be used without `Macro.escape` just for simple data structure like integer, string, ......

Issue #89, opened 2022-05-06
Create module parser


nappex:
Očekáváme předem známa vstupní data stažená downloaderem. Donwloader stahne textovy obsah stranky.
data tvořící text budou ve formě co řadek to jedna url (jako kdysi napč. gopher).
Vystupem parseru bude json soubor ktery bude obsahovat seznam vsech url nalezenutych v textu dané stranky.

nappex:
dalsi problemy

Jake formaty chceme nacitat.
Jake vystupni formaty chceme.

Glutexo:
#118 got merged; I am removing the “blocked” label.

Glutexo:
To answer your questions, @nappex:

The input of the parser will be anything downloaded from the Internet. Typically, this is an HTML page, but it can be a plain text file or a JSON API response. It is Spider’s job to know and discern the format; Onigumo can only provide a set of tools (like #87) to make everyday tasks easier.

The output format will be a structured representation of the input data. JSON was my first pick for its universality, and I preferred it over  YAML because the data will be processed only by a machine. The output can be, however, whatever kind of structured data, be it a serialized object of a programming language of the Spider, an XML document, or some binary format.

Theoretically, Onigumo doesn’t even need to force a concrete structure format. Having one would make it easy to define specific standardized nodes in the schema, allowing to control the core without writing any code. On the other hand, a piece of Spider code that knows its file format can provide the same data just as fine.

Feel free to share any thoughts on the topic!

Glutexo:
The work has begun: #146

Glutexo:
I extracted the steps described in your description, @nappex, to a new follow-up Issue #169. Feel free to take a look!

PR #145, opened 2022-08-23
✨ Use the first argument as a module name

Glutexo:
The first CLI argument with a module name enables the writing of runnable modules. A future _Parser_ module (#89) only needs a _main_ method taking the root path as an argument.

The dynamic approach also allows testing of the CLI entry point.

The approach is naïve. It does check neither the parameter count nor whether a module exists. A missing argument or an incorrect module name, including a different case, causes an uncaught error.

With the new code in place, the escript has a new invocation syntax:

```sh
$ ./onigumo Downloader
```


nappex:
I've got an error:

```
$ ./onigumo Downloader
** (HTTPoison.Error) :nxdomain
    (httpoison 1.8.0) lib/httpoison.ex:156: HTTPoison.request!/5
    (onigumo 0.1.0) lib/onigumo_downloader.ex:24: Onigumo.Downloader.download_url/2
    (elixir 1.12.2) lib/stream.ex:572: anonymous fn/4 in Stream.map/2
    (elixir 1.12.2) lib/stream.ex:1559: Stream.do_element_resource/6
    (elixir 1.12.2) lib/stream.ex:1719: Enumerable.Stream.do_each/4
    (elixir 1.12.2) lib/stream.ex:649: Stream.run/1
    (elixir 1.12.2) lib/kernel/cli.ex:124: anonymous fn/3 in Kernel.CLI.exec_fun/2
```

I invoked these commands

```
$ mix escript
$ mix escript.build
$ mix compile
```

nappex:
Tried to find something about :nxdomain on official doc, unfortunately without success.

`nxdomain error site:https://hexdocs.pm/httpoison` -> googled

nappex:
What is the `NXDOMAIN` error

When we try to enter a web page and we get an error message indicating NXDOMAIN, it means that the domain name could not be resolved. This logically means that we cannot navigate through that site we are trying to access.

nappex:
I've tried to debug `HTTPoison.get!` in terminal (iex) according to tutorial on official [doc](https://hexdocs.pm/httpoison/readme.html#usage).

Unfortunately I've got an error:

```
iex(1)> HTTPoison.start()
** (UndefinedFunctionError) function HTTPoison.start/0 is undefined (module HTTPoison is not available)
    HTTPoison.start()
```

I feel helplessly... Oh no my baby woke up seize

nappex:
If you want run
`iex(1)> HTTPoison.start()`

you have to run iex as:
`$ iex -S mix`

nappex:
nxdomain error is something what is expected because we use as input urls with domain `onigumo.local`, that domain does not exist in reality!

nappex:
We should consider to make special parser to parse arguments from cli to get module.

See https://elixirschool.com/en/lessons/intermediate/escripts
Then we can add more functionality to that parser, for example check numbers of args and so on.

nappex (approved):

Glutexo:
I agree that using _OptionParser_ is the correct way to process the arguments. I chose to use plain pattern matching only to keep the changeset to the lowest level possible and switch to the smarter approach in the follow-up pull request.

Also the argument is rudimentary in its nature. It is a raw module name. Passing an invalid one raises an exception. That too is a sign of a temporary simplistic solution.

Glutexo:
Thanks for your review! 🙌🏻 Merging.

Glutexo:
Thanks for mentioning `iex -S mix`. ❤️ I was surprised that there is no mix console command and there it is!

PR #146, opened 2022-08-26
Introduce a Parser module stub

Glutexo:
I created a dummy module to start working on the Parser component. It lists all files in the current working directory with no extension. In a directory that only contains files owned by Onigumo, such files are those fetched by the Downloader.

### DIFF
--- a/lib/onigumo.ex
+++ b/lib/onigumo.ex
@@ -57,3 +57,18 @@ defmodule Onigumo.Downloader do
     Application.get_env(:onigumo, :http_client)
   end
 end
+
+defmodule Onigumo.Parser do
+  @moduledoc """
+  Web scraper
+  """
+
+  def main(root_path) do
+    root_path
+    |> File.ls!()
+    |> Enum.reject(&File.dir?(&1))
+    |> Enum.reject(&String.contains?(&1, "."))
+    |> Enum.join("\n")
+    |> IO.puts()
+  end
+end


nappex:
I d like to add `Path.expand` check examples for `dir?` [hexdocs](hexdocs.pm/elixir/File.html#dir?/2). Its seems that can be problem with `~` in path.

lib/onigumo/parser.ex
```diff
@@ -0,0 +1,14 @@
+defmodule Onigumo.Parser do
+  @moduledoc """
+  Web scraper
+  """
+
+  def main(root_path) do
+    root_path
+    |> File.ls!()
+    |> Enum.reject(&File.dir?(&1))
+    |> Enum.reject(&String.contains?(&1, "."))
```
nappex, line 10: It means for hidden files or suffix?

Glutexo:
Currently, the root path is the current working directory, [emerging](https://github.com/Glutexo/onigumo/blob/10841b11cb1c9258bf53b9063926bc5a4a98ff95/lib/cli.ex#L4) in [Onigumo.CLI](https://github.com/Glutexo/onigumo/blob/10841b11cb1c9258bf53b9063926bc5a4a98ff95/lib/cli.ex). This value is always a real path that doesn’t require expansion. However, if the path becomes user-specified, it will need expanding by `Path.expand`.

Thank you for pointing this out! I created an issue #182 to track the possibility of setting the path via CLI.

lib/onigumo/parser.ex
Glutexo: Exactly. This removes all hidden files from the file listing. In addition, it removes `.` and `..` entries that are not real files.

Glutexo:
Merged curretn _master_ (https://github.com/Glutexo/onigumo/commit/dc0b3fc5ea24d0c5ed87815a48a5f93cb4f0d478).

Glutexo:
Thanks to pull request #147, downloaded files now have the .raw extension. This has simplified the process of loading downloaded files. Without the extension, it was more complicated.

Glutexo:
I’ve noticed that retrieving all files without an extension also includes the executable file `onigumo`.

Glutexo:
I introduced two changes:

- #147 changed the downloaded file names to have a _.raw_ extension. That resulted in a simpler and less error-prone file listing mechanism.
- _IO.puts_ does not print everything at once using _Enum.join_, but rather prints every item by itself using _Enum.map_.

lib/onigumo/parser.ex
```diff
@@ -7,7 +7,6 @@ defmodule Onigumo.Parser do
     root_path
     |> File.ls!()
 	|> Enum.filter(fn filename -> Path.extname(filename) == ".raw" end)
-    |> Enum.join("\n")
-    |> IO.puts()
+    |> Enum.map(&IO.puts(&1))
```
nappex: Interesting indentation. Is this indentation OK?

We can also use similar approach to python:
```
iex(4)> Path.absname("foo.raw") |> String.ends_with?(".raw")
true
```

lib/onigumo/parser.ex
```diff
@@ -0,0 +1,22 @@
...
+  	path
+    |> File.ls!()
+	|> Enum.filter(&is_downloaded(&1))
+  end
+
+  defp is_downloaded(path) do
+    suffix = Application.get_env(:onigumo, :downloaded_suffix)
+    Path.extname(path) == ".#{suffix}"
+  end
+end
```
nappex, line 22: I am confused by your indentations...

Glutexo: That was a mistake. Going to fix.

lib/onigumo/parser.ex
Glutexo: The indentation is not OK. I’ll fix it.

lib/onigumo/parser.ex
```diff
@@ -0,0 +1,26 @@
...
+    root_path
+    |> list_downloaded()
+    |> Enum.map(&IO.puts(&1))
+
+    :ok
+  end
+
+  defp list_downloaded(path) do
+    path
+    |> File.ls!()
```
nappex, line 18: Consider, if the variant with exclamation mark is necessary to use. Maybe the elixir way of handling error will be better approach with state `{:ok} or {:err, msg}` will be better.

lib/onigumo/parser.ex
```diff
@@ -0,0 +1,26 @@
...
+    |> Enum.filter(&is_downloaded(&1))
+  end
+
+  defp is_downloaded(path) do
+    suffix = Application.get_env(:onigumo, :downloaded_suffix)
```
nappex, line 23: The `donwloaded_suffix` should be constant during running of code, may the `compile_env` will be more suitable to ensure to use the value which is really specified in configuration and do not be aware that the value was somewhere changed during running of code.

lib/onigumo/parser.ex
Glutexo: I agree that we should avoid the bang functions. This is however a native function which we always expect to succeed. It is not our userspace function where we’d use the bang instead of a conditional.

We have #227 to test whether the path is valid. Now, we don’t do that even in the already existing Downloader: see the [`File.write!`](https://github.com/Glutexo/onigumo/blob/0d890b963ba18cca8d819bff58bd691adf27b458/lib/onigumo/downloader.ex#L43C14-L43C15) there.

With the path validation in place, any conditional here would imply that our validation may not work. I’d stick with the bang version then – even if it fails, it’d at least provide a useful error.

lib/onigumo/parser.ex
Glutexo: `:downloaded_suffix` is a (user-)configurable value. If it were compiled, it wouldn’t be possible to change it by editting the configuration file without recompilation. [get_env](https://hexdocs.pm/elixir/Application.html#get_env/3) makes sense here.

It’s not like [compile_env](https://hexdocs.pm/elixir/Application.html#get_env/3) in the module [mapping](https://github.com/Glutexo/onigumo/blob/0d890b963ba18cca8d819bff58bd691adf27b458/lib/cli.ex#L3) in the CLI module. It’s “configurable” only so it can be mocked in the test.

Glutexo:
In the light of #232, this module stub may look unnecessary. Yet it contains a draft of a file discovery logic, that can be re-used by the new Feeder module. Also, at this point, only tests are remaining. For the sake of continuity and code re-use, I’d finish this pull request as it is.

lib/onigumo/parser.ex
Glutexo: In Python, we can work with the string being path-aware too. [`os.path.splitext`](https://docs.python.org/ja/3.9/library/os.path.html#os.path.splitext) gets a filename extensions from a string path. I like to use a semantic function when there is one available, even though the path is a string.

lib/onigumo/parser.ex
Glutexo: I had a chance to write a few lines of code in Go. There, exceptions are even rarer than in Elixir. A function running into an error can cause “panic”. It can be rescued, but it’s definitely not good practice to do so for control flow. Like in Elixir, the functions return a tuple and you check for the error.

Moreover, I learned that file system operations can always fail. It may be rare or unexpected, but can happen for various reasons. Disk operations need to be treated in a similar way like e.g. HTTP requests.

We can (and should) implement a general rescue to handle exceptions. But we should not rely on it intentionally. That applies to cases when the error is unrescuable. We need to return and exit correctly, not letting an exception bubble up.

Thanks for your comment and insight!

nappex: I'd like to listen more details about it on Tuesday, especially about the filesystems. It sounds strange to me that we should implement general rescue to handle exceptions, if it is a best practice why does elixir not implement something like that as built-in?

PR #147, opened 2022-08-26
🚚 Append .raw to the downloaded file name

Glutexo:
As mentioned in https://github.com/Glutexo/onigumo/pull/146#issuecomment-1228710234, the files with an extension are more easily filterable by wildcards than files with none. It is also more consistent with having an extension for each file type.


Glutexo:
This is ready for review, @nappex.

test/onigumo_downloader_test.exs
```diff
@@ -116,7 +116,10 @@ defmodule OnigumoDownloaderTest do
       input_url = "https://onigumo.local/hello.html"
       created_file_name = Onigumo.Downloader.create_file_name(input_url)
 
-      expected_file_name = Onigumo.Utilities.Hash.md5(input_url, :hex)
+      input_url_hash = Onigumo.Utilities.Hash.md5(input_url, :hex)
+      downloaded_suffix = Application.get_env(:onigumo, :downloaded_suffix)
```
nappex, line 120: I don't like things where I have to get some value whose key is same as name of variable where value is saved. It is not a fancy way how to do it in my opinion, unfortunately I dont know how to do it better way and same problem I am not able to solve when I am programming in Python... That's just a note, because I'd like to discover if can be better approach or not... I am thinking about this kind of code often...

nappex (dismissed):

test/onigumo_downloader_test.exs
Glutexo: You're spot on about the potential confusion. In this specific test, the variable could simply be named suffix, as we're only using one. While the downloader_ prefix is useful in the configuration where multiple suffixes might exist, it could be misleading in our test code by implying we're using more than one suffix. Clearer naming will certainly help avoid such misunderstandings.

nappex (approved):

Issue #150, opened 2022-10-29
Use fucntion pattern matching to solve naming issues


nappex:
If we use function pattern matching feature available in elixir then we can use same name for some functions in `Downloader` module.

Consider to rename `main` as `download` and `download_urls_from_file` also as `download` then these will diff just by defined arguments. Download without any args will behave as our main and download with defined arg as only one a needs to be file will behave as our download_urls_from_file. This approach will solve our naming issue.

PR #151, opened 2022-10-29
Use download functions with pattern match

nappex:
This is just **proposal** for #150 ,

There is a problem where to write line `http_client.start()`, because now it is calling with each URL which should be handled.....

Maybe the Pattern matching should be more specific... not just by number of args but also by specific types for example...

### DIFF
--- a/lib/onigumo_downloader.ex
+++ b/lib/onigumo_downloader.ex
@@ -3,14 +3,14 @@ defmodule Onigumo.Downloader do
   Web scraper
   """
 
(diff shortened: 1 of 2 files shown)


nappex:
Some time ago was discussed that we dont want to use this func pattern matching. Check the closed RPs history.

Currently, we revised again and we decided to rename func `download_urls_from_file` because this func dont do any downloading, actually. It just prepares links saved from file and give to download_url. Renaming was done in #154

PR #152, opened 2022-10-29
Move Stream.run()

nappex:
The code seems more clear if we move `stream.run` to `download_urls_from_file` in my opinion


nappex:
Maybe it seems more clear but, then the code become more unpractical, we can not use download_urls_from_file as separate func imported from `Downloader`, If we wanna use in another module or in another stream....

PR #154, opened 2022-11-19
♻️ Rename download_urls_from_file

Glutexo:
The function does not download anything. It only creates a stream designated for downloading URL. This gives a false sense of complementarity to the download_url. The two functions however do very different things. Hence the rename.

### DIFF
--- a/test/onigumo_downloader_test.exs
+++ b/test/onigumo_downloader_test.exs
@@ -26,7 +26,7 @@ defmodule OnigumoDownloaderTest do
     end
   end
 
-  describe("Onigumo.Downloader.download_urls_from_file/1") do
+  describe("Onigumo.Downloader.create_download_stream/1") do
     @tag :tmp_dir
     test("download URLs from the input file", %{tmp_dir: tmp_dir}) do
       expect(HTTPoisonMock, :get!, length(@urls), &prepare_response/1)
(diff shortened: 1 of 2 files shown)


test/onigumo_downloader_test.exs
```diff
@@ -26,17 +26,17 @@ defmodule OnigumoDownloaderTest do
     end
   end
 
-  describe("Onigumo.Downloader.download_urls_from_file/1") do
+  describe("Onigumo.Downloader.create_download_stream/1") do
     @tag :tmp_dir
-    test("download URLs from the input file", %{tmp_dir: tmp_dir}) do
+    test("create a stream to download URLs from the input file", %{tmp_dir: tmp_dir}) do
```
nappex, line 31:
```suggestion
    test("download URLs from the input file with a created stream", %{tmp_dir: tmp_dir}) do
```

nappex (approved):

test/onigumo_downloader_test.exs
Glutexo: Thank you for the suggestion! I find it good, because this test does not focus on creating the stream, but on using the stream to download the URLs. Although without the function returning a stream, this test would fail on `|> Stream.run()`.

A proper unit test would verify that the stream is composed properly, comparing the actual stream to an expected one. There is a follow-up issue for that: #173.

PR #156, opened 2023-01-14
Update README - subgraphs names with Kernel and Application

nappex:


Glutexo (changes requested):
Unfortunately the chart does not render properly. 
スクリーンショット 2023-01-14 19 59 49

Glutexo (approved):

PR #159, opened 2023-01-15
Add test to invoke onigumo from CLI

### LINKED ISSUE #158: Test Onigumo CLI
Add some basic test to find out that commands `$ ./onigumo` or `$ ./onigumo Downloader` are running without any errors.

nappex:
fix #158

We dont test invocation Onigumo from CLI at this moment.
We should add some basic test to test that commands `$ ./onigumo` or `$ ./onigumo Downloader` in terminal are running without an error.

Unfortunately I have no idea how should be tested... So the PR is just draft to discuss..


Glutexo:
There are two options, not mutually exclusive:

- An integration test that would run the actual Downloader, only with HTTPoison mocked.
- A unit test that would test the CLI module but wouldn’t use the actual Downloader.

I think both are good to have. For the latter, we’d need to specify a [Behaviour](https://hexdocs.pm/elixir/Behaviour.html). That is not only a good learning opportunity but a correct thing to do too. The Downloader is a module with a public interface: the [_main_](https://github.com/nappex/onigumo/blob/301fd250e078d0d51b1fdddf298c0e6615b189fb/lib/onigumo_downloader.ex#L6) function.

test/onigumo_cli_test.exs
```diff
@@ -0,0 +1,11 @@
+defmodule OnigumoCLITest do
+  use ExUnit.Case
+
+
+  describe("Onigumo.CLI.main/1") do
+    test("run Onigumo.CLI.main") do
+      Onigumo.CLI.main(["arg1"])
```
Glutexo, line 7:
```suggestion
      Onigumo.CLI.main(["Downloader"])
```

Glutexo (changes requested):
The integration test will be almost identical to the existing “main” function test:

1. Prepare a list of URLs to download.
2. Invoke the Downloader.
3. Check the downloaded files.

A technical note: keep only the integration tests in this PR. For the future unit test, extract the behaviour introduction and the actual test into separate PRs.

test/onigumo_cli_test.exs
Glutexo: #145 got merged, the latter is now the correct call.

Glutexo:
@nappex, is everything clear? Do you have all the information you need to complete this pull request? Can I provide any help?

test/onigumo_cli_test.exs
Glutexo, line 7:
```suggestion
      Onigumo.CLI.main(["Downloader"])
```

Glutexo (commented):
If I recall correctly, you addressed this during our last meeting and had written an integration test for it, right? If that's the case, let's make sure to finalize it. We need that test in place. Let me know if you need any help or if there's anything I can do to assist you.

mix.exs
```diff
@@ -37,4 +38,8 @@ defmodule Onigumo.MixProject do
       main_module: Onigumo.CLI
     ]
   end
+
+  defp elixirc_paths(:test), do: ["lib", "test/support"]
+  defp elixirc_paths(_), do: [ "lib" ]
```
Glutexo, line 43:
```suggestion
  defp elixirc_paths(_), do: ["lib"]
```

Glutexo (changes requested):

nappex:
It is ready to merge.

There are two not ideal approaches:
1. we change `cwd`, because the `root_path` is set as constant and code does not see created files needed to further processes. For this case was issue created #184 so it should be fix by another PR.
2. This PR does not test invalid inputs to CLI. There is an issue for that case #183

test/support/input.ex
```diff
@@ -0,0 +1,6 @@
+defmodule InputSupport do
+  def prepare(urls) do
```
Glutexo, line 2: Maybe now when this is in a separate module, “prepare” is to vague. ”prepare_url_list“, “url_list”, “prepare_downloader_input”, “downloader_input”…

test/support/http.ex
```diff
@@ -0,0 +1,12 @@
+defmodule HttpSupport do
+  def prepare_response(url) do
```
Glutexo, line 2: Do we need that “prepare” here? The “body” function only describes the result with the word “body”, so the response function can be also named just “response”, doesn’t it?

Glutexo (dismissed):
💥

mix.exs
```diff
@@ -8,7 +8,8 @@ defmodule Onigumo.MixProject do
       elixir: "~> 1.10",
       start_permanent: Mix.env() == :prod,
       deps: deps(),
-      escript: escript()
+      escript: escript(),
+      elixirc_paths: elixirc_paths(Mix.env())
```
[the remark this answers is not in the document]
nappex: Good point we should look at closer!

test/support/http.ex
nappex: No we dont fixed, thanks

test/support/input.ex
nappex: Absolutely, thanks

mix.exs
Glutexo, line 12: `Mix.env()` is called twice. We can save the repeated call by storing the returned value to a variable.

Glutexo (approved):

PR #175, opened 2023-03-11
Fix ignoring the onigumo escript file

nappex:
- old definition ignore all occurency of onigumo in path
    - it is needed to explicitly speficied we want to ignore onigumo in
      root path

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,4 +25,5 @@ onigumo-*.tar
 # Temporary files, for example, from tests.
 /tmp/
 
-onigumo
+# Ignore onigumo escript file
+/onigumo


.gitignore
```diff
@@ -25,4 +25,5 @@ onigumo-*.tar
 # Temporary files, for example, from tests.
 /tmp/
 
-onigumo
+# Ignore onigumo escript file
```
Glutexo, line 28: This would theoratically belong into a different pull request. 😃 The fact that the line doesn’t have a comment is not related to the actual bug in the definitilon. Sorry for nitpicking.

Glutexo (approved):
Works like a charm! Thank you! ❤️

PR #177, opened 2023-03-11
Improve module naming and directory structure

### LINKED ISSUE #163: ♻️ Improve module naming and directory structure
All things Onigumo need to be inside the _Onigumo_ namespace to avoid name clashes. The Mix [project](https://github.com/Glutexo/onigumo/blob/076353f9bc5f7511181b2a5990ea2e75cd624715/mix.exs) itself is _Onigumo.MixProject_. **No submodules should pollute the global namespace.** That results in the following steps:

- [ ] Move the [_Downloader_](https://github.com/Glutexo/onigumo/blob/076353f9bc5f7511181b2a5990ea2e75cd624715/lib/onigumo_downloader.ex) module under a new _Kernel_ namespace.
- [ ] Move the [_Hash_](https://github.com/Glutexo/onigumo/blob/076353f9bc5f7511181b2a5990ea2e75cd624715/lib/hash.ex) module too into Onigumo.Kernel.
- [ ] Put the [_HTML_](https://github.com/nappex/onigumo/blob/af9da826466bb47486189b5330368350410c41a4/lib/spider_html.ex) module from #87 into _Onigumo.Spider_.

nappex:
fix #163 

Kernel is in `onigumo`. So we keep `Onigumo.Downloader` naming
Built-in spider modules are saved in `spider`. -> `Onigumo.Spider.HTML`
hash func as helping func of onigumo is moved to utilities, now. -> `Onigumo.Utilities.Hash`


Glutexo (approved):
I wondered, why the _Onigumo.CLI_ module is directly in _/lib_ and the _Onigumo.Downloader_ in the _/lib/onigumo_ subfolder. If I recall our discussion correctly, we came to a conclusion that the downloader, operator and parser are the main pillars that are low in count, so they don’t pollute the _Onigumo_ namespace much. Putting them under a _Onigumo_ or _Kernel_ namespace would result in overy long names like _Onigumo.Kernel.Downloader_, and _Onigumo.Onigumo_ looks silly.

The structure looks reasonable and it’s not chiseled in stone. I approve!

PR #186, opened 2023-06-09
Parse switches from CLI

nappex:
Parse switches for work directory. If we want to parse work dir we have parse newly component also.


nappex:
I am not sure about naming the switches, `output` for working directory is not something obvious at first sight. `Working` means where `inputs` are saved or `output` should going to or `both` at one place.

nappex:
Other task what should be done are tests.
I have to more explore the ways how to test private functions.

nappex:
We should consider library [Optimus](https://github.com/funbox/optimus) for parsing cli arguments as far as the onigumo is CLI app.

lib/cli.ex
```diff
@@ -1,7 +1,28 @@
...
-    module.main(root_path)
+  def main(argv) do
+    {parsed_switches, _, _} =
+      argv
+      |> parse_argv
+
+    module =
+      parsed_switches
+      |> Keyword.get(:component, "Downloader")
+      |> safe_concat()
```
nappex, line 10: We save the result to variable with name `module`, but the switch to be parse is called `component`. Would not be fancier when switch would be called as `module` also?

nappex: The variable name `module` is really module -> `Onigumo.Donwloader` is module and it is the thing we get from pipes to `module` variable.

Glutexo (changes requested):
We should split the pull request into two separate ones:

- [ ] Switch to _OptionParser_. -> #189 
- [ ] The new _output_ argument.

Also, _output_ may not be the best name for the argument. It is more a “working directory” rather than an output. The actual “output” is what the Materializer spits out.

Glutexo (commented):
❤️

nappex:
- [ ] Test working dir exist and contains corresponding files
- [ ] Test unreachable path, there is a file with same name with working-dir

Glutexo:
> * [ ]  Test working dir exist and contains corresponding files
> * [ ]  Test unreachable path, there is a file with same name with working-dir

For those, issues #218 and #219 were created.

PR #189, opened 2023-10-03
Use OptionParser to parse component

### LINKED ISSUE #162: Parse args from CLI with `OptionParser.parse`
Currently, we parse CLI args and options just with pattern matching #145. It will be nice if we parse args from CLI by built-in module `OptionParse` check [official docs](https://hexdocs.pm/elixir/OptionParser.html#parse/2) or [tutorial - elixir school](https://elixirschool.com/en/lessons/intermediate/escripts#parsing-args-1). This approach ensure to handle errors more elegant with help message and so on. Currently we get some ugly error. It is not very user friendly interface.

### LINKED ISSUE #188: Use CLI `component` as positional argument
We could consider updating the CLI usage from:

`./onigumo --component Downloader --working-dir ~/wd/`

to

`./onigumo Downloader --working-dir ~/wd/`

nappex:
fix #162 
fix #188 

I'be tried to be really consistent with existing code and make really the smallest diff as possible.
That's the one of reasons why I parse the component as positional arguments.

Nice thing was appeared, the `OptionParser` is not so good in parsing the positional arguments and its main goal is parsing `switches` (resp. `options`).

IMHO it is emphased by warning which you get from elixir, when you do not pass `:switches` or `:strict`, but these are required only for `switches`.

```shell
warning: not passing the :switches or :strict option to OptionParser is deprecate
```

### DIFF
--- a/test/onigumo_cli_test.exs
+++ b/test/onigumo_cli_test.exs
@@ -32,5 +32,9 @@ defmodule OnigumoCLITest do
     test("run CLI with more than one argument") do
       assert_raise(MatchError, fn -> Onigumo.CLI.main(["Downloader", "Parser"]) end)
     end
+
+    test("run CLI with invalid switch") do
+      assert_raise(OptionParser.ParseError, fn -> Onigumo.CLI.main(["--help"]) end)
+    end
   end
 end
--- a/test/onigumo_cli_test.exs
+++ b/test/onigumo_cli_test.exs
@@ -26,11 +26,11 @@ defmodule OnigumoCLITest do
     end
 
     test("run CLI with no arguments") do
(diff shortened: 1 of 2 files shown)


test/onigumo_cli_test.exs
```diff
@@ -21,16 +21,20 @@ defmodule OnigumoCLITest do
       Onigumo.CLI.main(["Downloader"])
     end
 
-    test("run CLI with unknown argument") do
+    test("run CLI with invalid argument") do
```
Glutexo, line 24: Maybe this is a separate change, but I see the change for consistency with the new “invalid switch” tests. It is okay to have it here.

Glutexo (approved):
Awesome! 🎉 Great work! :shipit:

PR #193, opened 2023-10-07
Use OptionParser.parse

### LINKED ISSUE #204: Use functions with trailing bang(!) or without?
https://elixir-br.github.io/getting-started/try-catch-and-rescue.html

nappex:
fix #204

### DIFF
--- a/lib/cli.ex
+++ b/lib/cli.ex
@@ -1,6 +1,7 @@
 defmodule Onigumo.CLI do
-  def main([component]) do
-    module = Module.safe_concat("Onigumo", component)
+  def main(argv) do
+    {_parsed, args, _invalid} = OptionParser.parse(argv, strict: [])
+
     root_path = File.cwd!()
     module.main(root_path)
   end
--- a/lib/cli.ex
+++ b/lib/cli.ex
@@ -2,7 +2,19 @@ defmodule Onigumo.CLI do
   def main(argv) do
     {_parsed, args, _invalid} = OptionParser.parse(argv, strict: [])
 
+    module =
+      args
+      |> get_component
+      |> safe_concat
+
     root_path = File.cwd!()
     module.main(root_path)
+  end
+
+  defp get_component([]), do: "Downloader"
+  defp get_component([component | _]), do: component
+
+  defp safe_concat(component) do
+    Module.safe_concat("Onigumo", component)
   end
 end


lib/cli.ex
```diff
@@ -1,7 +1,20 @@
 defmodule Onigumo.CLI do
-  def main([component]) do
-    module = Module.safe_concat("Onigumo", component)
+  def main(argv) do
+    {_parsed, args, _invalid} = OptionParser.parse(argv, strict: [])
```
Glutexo, line 3: If we match the invalid arguments with `[]`, we‘ll get a MatchError. Like that, we can leverage the current [rescue clause](https://github.com/Glutexo/onigumo/blob/f7272069fcc32abdc4ddc6333fae150d657c0f08/lib/cli.ex#L10C5-L10C11) catching for _OptionParser.ParseError_ and _MatchError_. It’s not something we want to keep as we want to abandon raising exceptions (see #204), but it can be a nice step that keeps a minimal amount of changes.

The same applies for the parsed argument, which we [already do](https://github.com/Glutexo/onigumo/blob/f7272069fcc32abdc4ddc6333fae150d657c0f08/lib/cli.ex#L8) in the codebase. They both would raise a _MatchError_ and we currently don’t need to differentiate between them.

```suggestion
    {[], args, []} = OptionParser.parse(argv, strict: [])
```

lib/cli.ex
```diff
@@ -1,7 +1,20 @@
...
+    module =
+      args
+      |> get_component
+      |> safe_concat
+
     root_path = File.cwd!()
     module.main(root_path)
   end
+
+  defp get_component([]), do: "Downloader"
```
Glutexo, line 14: This adds additional logic of setting a default component. That’s not necessarily wrong, at least now when there is only a single component, but it’s over the scope of introducing _OptionParser_.

lib/cli.ex
```diff
@@ -1,7 +1,20 @@
 defmodule Onigumo.CLI do
-  def main([component]) do
-    module = Module.safe_concat("Onigumo", component)
+  def main(argv) do
+    {_parsed, args, _invalid} = OptionParser.parse(argv, strict: [])
+
+    module =
+      args
+      |> get_component
+      |> safe_concat
```
Glutexo, line 8: After a productive discussion, we abandoned _safe_concat_ in favor of a [module literal](https://github.com/Glutexo/onigumo/blob/f7272069fcc32abdc4ddc6333fae150d657c0f08/lib/cli.ex#L3).

lib/cli.ex
```diff
@@ -1,7 +1,20 @@
...
     root_path = File.cwd!()
     module.main(root_path)
   end
+
+  defp get_component([]), do: "Downloader"
+  defp get_component([component | _]), do: component
+
+  defp safe_concat(component) do
+    Module.safe_concat("Onigumo", component)
+  end
```
Glutexo, line 19: As mentioned above. Thanks to abandoning _safe_concat_ as unsafe 😃, exposing new errors and abusing reflection, we won’t need this function altogether.

```suggestion
```

Glutexo (changes requested):
Also, there are no tests. But I know this is only a draft.

lib/cli.ex
```diff
@@ -1,14 +1,14 @@
 defmodule Onigumo.CLI do
-  @components %{
+ @components %{
```
Glutexo:
```suggestion
  @components %{
```

lib/cli.ex
```diff
@@ -5,10 +5,10 @@ defmodule Onigumo.CLI do
 
   def main(argv) do
     try do
-      {[], [component]} = OptionParser.parse!(argv, strict: [])
+      {[], [component], []} = OptionParser.parse(argv, strict: [])
```
Glutexo, line 8: Just mentioning that in this case, there are now two different MatchErrors that can occur:

- An invalid switch is provided, causing the third assignment `[]` to not match.
- More than one positional argument causing the secund assignment `[component]` to fail.

We don’t, however, need to differentiate these two, as both mean invalid CLI usage.

Glutexo (approved):

Glutexo (commented):
Fix the whitespace.

Glutexo (changes requested):
Fix the whitespace.

Glutexo (approved):
Sweet!

PR #195, opened 2023-10-15
Parse working_dir switch

### LINKED ISSUE #182: Specify root path (--working-dir) by CLI argument

nappex:
fix #182


lib/cli.ex
```diff
@@ -1,8 +1,10 @@
...
   def main(argv) do
-    {[], [component]} = OptionParser.parse!(argv, strict: [])
+    {parsed_switches, [component]} = OptionParser.parse!(argv, strict: [working_dir: :string])
     module = Module.safe_concat("Onigumo", component)
-    root_path = File.cwd!()
-    module.main(root_path)
+
+    parsed_switches
+    |> Keyword.get(:working_dir, File.cwd!())
+    |> module.main
```
Glutexo, line 8: The Elixir style guide [suggests](https://github.com/christopheradams/elixir_style_guide?tab=readme-ov-file#parentheses-pipe-operator) using parentheses in the pipe operator.

> Use parentheses for one-arity functions when using the pipe operator (`|>`).

```suggestion
    |> module.main()
```

I’d generally incline towards using them for functions without an argument, mainly for consistency. The Guide considers that [preferred](https://github.com/christopheradams/elixir_style_guide?tab=readme-ov-file#function-calls-and-parentheses), although in other examples they use them without parens a lot themselves.

> Use parentheses in function calls, especially inside a pipeline.

Glutexo (changes requested):

nappex:
DONT FORGET TO TESTS!! ⚠️

Glutexo:
Test lowercase `-c` too.

Glutexo (commented):
Works like a charm! ❤️  Now it’s time to write tests. 💪🏻

nappex:
Reade for review I've already merged #223 to this PR!

lib/cli.ex
```diff
@@ -4,11 +4,15 @@ defmodule Onigumo.CLI do
   }
 
   def main(argv) do
-    case OptionParser.parse(argv, strict: []) do
-      {[], [component], []} ->
+    case OptionParser.parse(
+           argv,
+           aliases: [C: :working_dir],
```
Glutexo, line 9: Did we agree on including the alias directly and not create a separate follow-up PR?

lib/cli.ex
```diff
@@ -4,11 +4,15 @@ defmodule Onigumo.CLI do
...
+           strict: [working_dir: :string]
+         ) do
+      {parsed_switches, [component], []} ->
         {:ok, module} = Map.fetch(@components, String.to_atom(component))
-        root_path = File.cwd!()
-        module.main(root_path)
+        working_dir = Keyword.get(parsed_switches, :working_dir, File.cwd!())
```
Glutexo, line 14: _root_path_ has been renamed to working_dir to correspond with the switch name, right? _working_dir_ is definitely a better name, _root_path_ may be confusing.

lib/cli.ex
```diff
@@ -19,11 +23,14 @@ defmodule Onigumo.CLI do
...
     IO.puts("""
-    Usage: onigumo [COMPONENT]
+    Usage: onigumo [OPTION]... [COMPONENT]
 
     Simple program that retrieves HTTP web content as structured data.
 
     COMPONENT\tOnigumo component to run, available: #{components}
+
+    OPTIONS:
+    -C, --working-dir=DIR\tChange to <DIR> before processing remaining files
```
Glutexo: The short version does not use `=`, but the long one may. It may be confusing to use it in the usage example, because it makes the impression that `-C=<dir>` is a valid usage. However, wget does it exactly like this and apparently it’s not a problem. 😃

cURL, on the other hand, does not use `=`, but a space. It also uses `<lowercase>` notation for the variables and does not mention the variable in the description. Like this:

```suggestion
    -C, --working-dir <dir>\tChange working directory before processing remaining files
```

For completeness, _tar_ has it like this:

```
-C <dir>  Change to <dir> before processing remaining files
```

I don’t have any strong preference. Just mentioning.

Glutexo: Change _what_? What about “change the (current) working directory”. I know that tar uses your wording, but tar is generally confusing. 🤯

What do you mean by “before processing remaining files”? This is exactly what tar says, but I am not sure it’s appropriate in our context. Even though Onigumo in fact processes files, just as any software does, I wouldn’t call it that way if I was talking about it. Moreover, the “remaining” there sounds confusing too: it would make more sense in case of resumed run, but, again, it’s not the primary use case. Let’s say something more universal and straightforward like Meson’s “before running”.

```suggestion
    -C, --working-dir=DIR\tChange the working directory before running
```

lib/cli.ex
nappex: I suppose yes and we agree that the change is so small (one line in parser and two chars in usage message) that it is not necessary to make a whole new PR for that.

lib/cli.ex
nappex: 🤙

Glutexo: I found my issue #222 where I suggested the rename. Linked.

lib/cli.ex
nappex: I prefer the way how the cURL do it, because my experience that typing with **space** is more common in reality then **=**.

nappex: Alright that is good point I've changed to running.

test/onigumo_cli_test.exs
```diff
@@ -23,8 +28,10 @@ defmodule OnigumoCLITest do
       assert usage_message_printed?(fn -> Onigumo.CLI.main(["Downloader", "Parser"]) end)
     end
 
-    test("run CLI with invalid switch") do
-      assert usage_message_printed?(fn -> Onigumo.CLI.main(["--invalid"]) end)
+    for switch <- @invalid_switches do
+      test("run CLI with invalid switch #{switch}") do
+        assert usage_message_printed?(fn -> Onigumo.CLI.main([unquote(switch)]) end)
+      end
```
Glutexo, line 34: We don’t use exit codes! 😱  It struck me when I saw them not being tested. Not for this PR, but created an issue #224.

test/onigumo_cli_test.exs
```diff
@@ -35,6 +42,20 @@ defmodule OnigumoCLITest do
...
+
+      assert Onigumo.CLI.main(["downloader", "--working-dir", tmp_dir]) == tmp_dir
+    end
+
+    @tag :tmp_dir
+    test("run CLI 'downloader' with '-C' switch", %{tmp_dir: tmp_dir}) do
+      expect(OnigumoDownloaderMock, :main, fn root_path -> root_path end)
+
+      assert Onigumo.CLI.main(["downloader", "-C", tmp_dir]) == tmp_dir
+    end
```
Glutexo, line 57: Could this be parametrized, using for with a list of valid arguments. Not only it’s the same test repeated twice, but also it’s the same switch, just with an alias.

```suggestion
    @working_dir_switches ["--working-dir", "-C"]

    @tag :tmp_dir
    for switch <- @working_dir_switches do
      test("run CLI 'downloader' with #{inspect(switch)} switch", %{tmp_dir: tmp_dir}) do
        expect(OnigumoDownloaderMock, :main, fn root_path -> root_path end)

        assert Onigumo.CLI.main(["downloader", switch, tmp_dir]) == tmp_dir
      end
```

test/onigumo_cli_test.exs
```diff
@@ -23,8 +28,10 @@ defmodule OnigumoCLITest do
       assert usage_message_printed?(fn -> Onigumo.CLI.main(["Downloader", "Parser"]) end)
     end
 
-    test("run CLI with invalid switch") do
-      assert usage_message_printed?(fn -> Onigumo.CLI.main(["--invalid"]) end)
+    for switch <- @invalid_switches do
+      test("run CLI with invalid switch #{switch}") do
```
Glutexo, line 32: Let’s use `inspect` for consistency. It makes it clear that’s an inserted variable.

```suggestion
      test("run CLI with invalid switch #{inspect(switch)}") do
```

test/onigumo_cli_test.exs
```diff
@@ -35,6 +42,20 @@ defmodule OnigumoCLITest do
       assert Onigumo.CLI.main(["downloader"]) == tmp_dir
     end
 
+    @tag :tmp_dir
+    test("run CLI 'downloader' with '--working-dir' switch", %{tmp_dir: tmp_dir}) do
+      expect(OnigumoDownloaderMock, :main, fn root_path -> root_path end)
```
Glutexo, line 47: Let’s be consistent and stick to the rename.

```suggestion
      expect(OnigumoDownloaderMock, :main, fn working_dir -> working_dir end)
```

test/onigumo_cli_test.exs
```diff
@@ -35,6 +42,20 @@ defmodule OnigumoCLITest do
...
+
+      assert Onigumo.CLI.main(["downloader", "--working-dir", tmp_dir]) == tmp_dir
+    end
+
+    @tag :tmp_dir
+    test("run CLI 'downloader' with '-C' switch", %{tmp_dir: tmp_dir}) do
+      expect(OnigumoDownloaderMock, :main, fn root_path -> root_path end)
```
Glutexo, line 54: dtto

```suggestion
      expect(OnigumoDownloaderMock, :main, fn working_dir -> working_dir end)
```

test/onigumo_cli_test.exs
```diff
@@ -35,6 +42,20 @@ defmodule OnigumoCLITest do
...
+
+      assert Onigumo.CLI.main(["downloader", "-C", tmp_dir]) == tmp_dir
+    end
+
```
Glutexo, line 58: I am missing a test verifying what happens if you pass in the `--working-dir`/`-C` switch without a value. Like

```
$ ./onigumo downloader --working-dir
```

Does it use the default, or does it pass an empty string? Is it a valid use case?

lib/cli.ex
Glutexo, line 14: What happens if there is no value passed to the switch? Like

```
./onigumo downloader --working-dir
```

Glutexo (changes requested):

lib/cli.ex
```diff
@@ -19,11 +23,14 @@ defmodule Onigumo.CLI do
...
     IO.puts("""
-    Usage: onigumo [COMPONENT]
+    Usage: onigumo [OPTION]... [COMPONENT]
 
     Simple program that retrieves HTTP web content as structured data.
 
     COMPONENT\tOnigumo component to run, available: #{components}
+
+    OPTIONS:
+    -C, --working-dir <dir>\tChange to <dir> before running
```
Glutexo, line 33:
```suggestion
    -C, --working-dir <dir>\tChange working dir to <dir> before running
```

test/onigumo_cli_test.exs
```diff
@@ -8,6 +8,11 @@ defmodule OnigumoCLITest do
     "uploader"
   ]
 
+  @invalid_switches [
+    "--invalid",
+    "-c"
+  ]
```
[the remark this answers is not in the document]
nappex: Frankly I do not know the initial intention... I just wanted to be consistent with already exist code.

test/onigumo_cli_test.exs
nappex: Absolutely, thank you.

test/onigumo_cli_test.exs
```diff
@@ -23,18 +33,30 @@ defmodule OnigumoCLITest do
...
       assert Onigumo.CLI.main(["downloader"]) == tmp_dir
     end
 
+
+    for switch <- @working_dir_switches do
+      @tag :tmp_dir
+      test("run CLI 'downloader' with #{inspect(switch)} switch", %{tmp_dir: tmp_dir}) do
+        expect(OnigumoDownloaderMock, :main, fn working_dir -> working_dir end)
+
+        assert Onigumo.CLI.main(["downloader", unquote(switch), tmp_dir]) == tmp_dir
```
Glutexo, line 56: It’s nice what an ingenious trick we found to check the actual value of the working dir. Maybe it’s time to create a separate function to parse the CLI options. Not in this PR, though.

test/onigumo_cli_test.exs
```diff
@@ -23,18 +33,36 @@ defmodule OnigumoCLITest do
...
     @tag :tmp_dir
     test("run CLI with 'downloader' argument passing cwd", %{tmp_dir: tmp_dir}) do
-      expect(OnigumoDownloaderMock, :main, fn root_path -> root_path end)
+      expect(OnigumoDownloaderMock, :main, fn working_dir -> working_dir end)
 
       File.cd(tmp_dir)
       assert Onigumo.CLI.main(["downloader"]) == tmp_dir
     end
 
+
```
Glutexo, line 50:
```suggestion

```

Glutexo (changes requested):

test/onigumo_cli_test.exs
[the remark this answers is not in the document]
Glutexo: And it may have even be myself who suggested the newlines. 😳

lib/cli.ex
nappex: Test added.

test/onigumo_cli_test.exs
```diff
@@ -23,18 +33,41 @@ defmodule OnigumoCLITest do
...
+      assert usage_message_printed?(fn -> Onigumo.CLI.main(["downloader", "--working-dir"]) end)
+    end
+
+    @tag :tmp_dir
+    test("run CLI 'downloader' with '--working-dir' and value '.'", %{tmp_dir: tmp_dir}) do
+      expect(OnigumoDownloaderMock, :main, fn working_dir -> working_dir end)
+
+      File.cd(tmp_dir)
+      assert Onigumo.CLI.main(["downloader", "--working-dir", "."]) == "."
+    end
```
Glutexo, line 69: I know it was me who came up with the idea of adding this test. After the discussion at the end of our call, I am no longer sure it tests anything useful. It only verifies that the dot it parsed as a value for the switch, which tests the [OptionParser](https://hexdocs.pm/elixir/OptionParser.html), not anything in Onigumo.

It should be the downloader, or more precisely – the component – or even more precisely – a middleware (It may be a use case for the [use](https://hexdocs.pm/elixir/alias-require-and-import.html#use) macro.) that would ensure proper handling of paths. But maybe [Optimus](https://github.com/funbox/optimus) could handle that, just as Python’s [argparse](https://docs.python.org/3.12/library/argparse.html) has its [FileType](https://docs.python.org/3.12/library/argparse.html#argparse.FileType).

I am sorry about that, but I’d vote for rather not having this test here. We can (and should) revisit path processing later. I won’t however block the PR if the test remains present.

Glutexo (changes requested):

test/onigumo_cli_test.exs
Glutexo: Update: [Optimus](https://github.com/funbox/optimus) doesn’t have a file type by itself. It supports custom parsers though, so a reusable one can be written. It is even possible that there already is a [Hex](https://hex.pm/) package for that. Although it [seems](https://hex.pm/packages?search=optimus&sort=recent_downloads) like there isn’t one. Maybe we could publish our work as a side-quest? 😱

Glutexo:
Just discovered a [Mix.Shell](https://hexdocs.pm/mix/Mix.Shell.html#c:cmd/2) module that has a [cmd](https://hexdocs.pm/mix/Mix.Shell.html) function with a `:cd` argument. I am not sure yet what’s that good for, but it looks related.

test/onigumo_cli_test.exs
```diff
@@ -23,18 +33,41 @@ defmodule OnigumoCLITest do
...
+      test("run CLI 'downloader' with #{inspect(switch)} switch", %{tmp_dir: tmp_dir}) do
+        expect(OnigumoDownloaderMock, :main, fn working_dir -> working_dir end)
+
+        assert Onigumo.CLI.main(["downloader", unquote(switch), tmp_dir]) == tmp_dir
+      end
+    end
+
+    test("run CLI 'downloader' with '--working-dir' without any value") do
+      assert usage_message_printed?(fn -> Onigumo.CLI.main(["downloader", "--working-dir"]) end)
+    end
```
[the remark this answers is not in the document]
nappex: Thanks for review, it is really valuable in this case.

test/onigumo_cli_test.exs
```diff
@@ -23,18 +33,41 @@ defmodule OnigumoCLITest do
...
+
+      test("run CLI 'downloader' with #{inspect(switch)} without any value") do
+        assert usage_message_printed?(fn -> Onigumo.CLI.main(["downloader", unquote(switch)]) end)
+      end
+
+      @tag :tmp_dir
+      test("run CLI 'downloader' with #{inspect(switch)} and value '.'", %{tmp_dir: tmp_dir}) do
+        expect(OnigumoDownloaderMock, :main, fn working_dir -> working_dir end)
+
+        File.cd(tmp_dir)
```
Glutexo, line 66: This doesn't do anything, actually. Since the Downloader is mocked, the working directory is not used for anything, not even in the tst.

```suggestion
        File.cd(tmp_dir)
```

Glutexo (approved):
🎉 

Thank you for the changes! ❤️

PR #212, opened 2024-02-24
Do not rescue MatchErrors in CLI

### LINKED ISSUE #211: Don’t rescue MatchErrors
The CLI module currently uses pattern matching that raises MatchError. That’s not a proper functional way of handling unexpected return values. We should rather capture any returned value and verify it subsequently without raising and rescuing an error.

nappex:
fix #211


Glutexo (approved):
🚀

PR #223, opened 2024-04-06
Remove unused @urls

Glutexo:
`@urls` module variable was only needed for the actual downloader workings. Now it being mocked, the acutual URLs are irrelevant.


Glutexo:
Thanks @nappex for spotting this!

nappex (approved):
It really seems that only usage for that was `HTTPoisonMock`

PR #238, opened 2024-08-16
Implement switch -h, --help to CLI

### LINKED ISSUE #205: Implement option `--help`, `-h`

nappex:
Fix #205

Blocked by #260.


Glutexo (commented):
Tests? 🤔

nappex:
> Tests? 🤔

thanks for reminding 😮‍💨

test/onigumo_cli_test.exs
```diff
@@ -60,6 +60,12 @@ defmodule OnigumoCLITest do
       end
     end
 
+    for switch <- ["-h", "--help", "--help -C prdel", "-h --invalid-switch"] do
```
Glutexo, line 63: I believe that there are two separate user stories:

* The help message is explicitly requested by providing the `-h`/`--help` switch.
* Invalid arguments are provided and the help message gets printed as a side effect.

I’d suggest then splitting this test into separate ones for each case.

```suggestion
    for switch <- ["-h", "--help"] do
    end
    …
    for switch <- ["--help -C prdel", "-h --invalid-switch"] do
    end
```

Glutexo (changes requested):

lib/cli.ex
```diff
@@ -30,6 +33,7 @@ defmodule Onigumo.CLI do
     COMPONENT\tOnigumo component to run, available: #{components}
 
     OPTIONS:
+    -h, --help\t\tusage help
```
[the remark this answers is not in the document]
Glutexo: Unfortunately, this hasn’t been addressed.

test/onigumo_cli_test.exs
```diff
@@ -60,6 +60,17 @@ defmodule OnigumoCLITest do
       end
     end
 
+    for switch <- ["-h", "--help"] do
+      test("run CLI with a #{inspect(switch)} switch") do
+        assert usage_message_printed?(fn -> Onigumo.CLI.main([unquote(switch)]) end)
+      end
+    end
+    for switches <- ["--help -C prdel", "-h --invalid-switch"] do
```
Glutexo:
```suggestion

    for switches <- ["--help -C prdel", "-h --invalid-switch"] do
```

test/onigumo_cli_test.exs
```diff
@@ -60,6 +60,17 @@ defmodule OnigumoCLITest do
...
+      test("run invalid combination of swiches #{inspect(switches)} ") do
```
Glutexo, line 70:
```suggestion
      test("run invalid combination of switches #{inspect(switches)} ") do
```

test/onigumo_cli_test.exs
```diff
@@ -60,6 +60,18 @@ defmodule OnigumoCLITest do
...
     end
 
+    for switch <- ["-h", "--help"] do
+      test("run CLI with a #{inspect(switch)} switch") do
+        assert usage_message_printed?(fn -> Onigumo.CLI.main([unquote(switch)]) end)
+      end
+    end
+
+    for switches <- ["--help -C invalid", "-h --invalid"] do
+      test("run invalid combination of swiches #{inspect(switches)} ") do
```
Glutexo, line 70: This may not be entirely exclusive use case for `-h`/`--help`, but rather a generic invalid combination of switches. I can imagine this being moved right after the existing invalid switch test, being different by providing a combination of a valid and an invalid argument.

I see even more user stories for the future:

* a valid and an invalid arguments provided at the same time,
* and an invalid combination of otherwise all valid arguments.

In our case, the latter is a combination of `-h`/`--help` and the component name.

Glutexo (changes requested):

lib/cli.ex
```diff
@@ -30,6 +33,7 @@ defmodule Onigumo.CLI do
     COMPONENT\tOnigumo component to run, available: #{components}
 
     OPTIONS:
+    -h, --help\t\tprint this help
```
Glutexo, line 36: Capitalized, this sentence would be more consistent.

```suggestion
    -h, --help\t\tPrint this help
```

test/onigumo_cli_test.exs
```diff
@@ -39,6 +44,12 @@ defmodule OnigumoCLITest do
       end
     end
 
+    for switches <- @invalid_switch_combinations do
```
Glutexo, line 47: There is one more user story of providing both a positional argument and the help switch. If ran in the `./onigumo downloader -h` order, the `-h` switch is ignored.

This case is not tested and because it doesn’t work correctly, the test would fail. The expected result is a help message to be printed because of an invalid argument combination. The actual result is that the downloader is run.

Glutexo (changes requested):
Unfortunately, a combination of `-h`/`--help` and a positional argument doesn’t work as expected. See my inline comment.

Glutexo:
My observation of `onigumo downloader -h` not printing the help message is not a matter of this pull request. It is a bug in the code preventing switches from being parsed if a component is provided. `onigumo downloader --anything` would behave identically.

The question is whether to add a test case for the combination of the help switch and the component. Because of the existing bug, the test would fail and essentially make this pull request blocked. We can keep this test for the fix and by that let this pull request go through.

Glutexo:
In my previous comment, I was wrong! 😮 It’s only the `-h`/`--help` switch that is ignored. An invalid switch like `-x` or `--invalid` still causes a help message to be printed.

```
$ ./onigumo downloader -h
```

```
$ ./onigumo downloader -x
Usage: onigumo [OPTION]... [COMPONENT]

COMPONENT	Onigumo component to run, available: downloader

OPTIONS:
-h, --help		print this help
-C, --working-dir <dir>	Change working dir to <dir> before running

```

Glutexo:
I found the reason, why the behavior changed:

* On [master](https://github.com/Glutexo/onigumo/tree/a4e350e053c953f0142fa6e6e3f13cd55118c97a), neither the `--help` switch` nor the `-h` alias were defined in `strict` or `aliases` respectively. Because of that, `--help`/`-h` ended up in the `invalid` section of the result, not matching the empty list in `{parsed_switches, [component], []}` and proceeding to `_` printing a help message.
* On [implement-help-option](https://github.com/nappex/onigumo/tree/3703cdbd2ef2faf92a710ac3dd7919b8eb10751b), however, the help switches are defined and because of that `:help` is not put in `invalid`. `[]` of invalid switches is then properly matched by `{parsed_switches, [component], []}` and the help switch ends up unparsed in `parsed_switches`.

This pull request then does introduce the bug in the sense that in didn’t exist because there was no help option. At the same time, it only reveals an existing issues of unscalable pattern matching of the `OptionParser.parse/2` output.

This can be still resolved tactically by using my code suggestion still being purely based on pattern-matching, probably in a separate pull request unblocking this one. The strategic solution is however to replace the pattern matching by something scalable.

test/onigumo_cli_test.exs
nappex: that should be already tested but it is strange that github does not mark as as outdated. I suppose it because the line of code you mark is not changed but in text is mentioned case of test which is not added in your changes.

nappex:
`--help` must be used as only one cli argument to run correctly, we do not want behaviour as "rule them all" by help switch as proposed in PR #244.
If someone use `--help` with other argument or switch, then it could evoke to user it should output help for certain argument resp. switch. We do not want to implement this feature currently, maybe later. So `--help` should be used only for general help message for whole `Onigumo` not for its individual commands.

Glutexo:
I think #256 would make this much easier to implement.

lib/cli.ex
```diff
@@ -4,16 +4,26 @@ defmodule Onigumo.CLI do
...
+        argv,
+        aliases: [h: :help, C: :working_dir],
+        strict: [help: :boolean, working_dir: :string]
+      )
 
-    with {switches, [component], []} <- parsed,
-         {:ok, module} <- Map.fetch(@components, String.to_atom(component)) do
-      working_dir = Keyword.get(switches, :working_dir, File.cwd!())
-      module.main(working_dir)
+    with {[help: true], [], []} <- parsed do
```
Glutexo, line 14: I am afraid this is not the right use case for `with`. It only collects a single value and the `else` part only contains a single branch. It’s just a pompous `if`/`else`. `case` might be more appropriate here. See  #256.

```elixir
case parsed:
  [help: true], [], []} ->
    usage_message()

  {switches, [component], []} ->
    …

  {_, _, [_ | _]} ->
    usage_message()

  {_, argv, _} when length(argv) != 1 →
    usage_message()
end
```

lib/cli.ex
```diff
@@ -4,16 +4,26 @@ defmodule Onigumo.CLI do
...
-      {_, _, [_ | _]} -> usage_message()
-      {_, argv, _} when length(argv) != 1 -> usage_message()
-      :error -> usage_message()
+      _ ->
+        with {switches, [component], []} <- parsed,
+             {:ok, module} <- Map.fetch(@components, String.to_atom(component)) do
+          working_dir = Keyword.get(switches, :working_dir, File.cwd!())
+          module.main(working_dir)
+        else
+          {_, _, [_ | _]} -> usage_message()
```
Glutexo, line 23: Unlike the other two patterns in this `else` clause, this one catches a case when any invalid switch is passed. It belongs more to the level where the help switch and the component run are matched.

```elixir
case parsed do
  {[help: true], [], []} ->
    …

  {switches, [component], []} ->
    …

  {_, _, [_ | _]} →
    …
end
```

lib/cli.ex
```diff
@@ -4,16 +4,26 @@ defmodule Onigumo.CLI do
...
+          {_, argv, _} when length(argv) != 1 -> usage_message()
```
Glutexo, line 24: This pattern is related to the component run, so it can belong here.

Glutexo (changes requested):

test/onigumo_cli_test.exs
```diff
@@ -13,6 +13,12 @@ defmodule OnigumoCLITest do
     "-c"
   ]
 
+  @invalid_switch_combinations [
```
Glutexo, line 16: These are not switch combinations. Switches and positional arguments are combined.

```suggestion
  @invalid_combinations [
```

test/onigumo_cli_test.exs
```diff
@@ -39,6 +45,12 @@ defmodule OnigumoCLITest do
       end
     end
 
+    for switches <- @invalid_switch_combinations do
```
Glutexo, line 48:
```suggestion
    for switches <- @invalid_combinations do
```

test/onigumo_cli_test.exs
```diff
@@ -39,6 +45,12 @@ defmodule OnigumoCLITest do
       end
     end
 
+    for switches <- @invalid_combinations do
```
Glutexo, line 48:
```suggestion
    for combination <- @invalid_combinations do
```

test/onigumo_cli_test.exs
```diff
@@ -39,6 +45,12 @@ defmodule OnigumoCLITest do
...
+      test("run invalid combination of switches #{inspect(switches)} ") do
```
Glutexo:
```suggestion
      test("run CLI with invalid combination #{inspect(combination)} ") do
```

test/onigumo_cli_test.exs
```diff
@@ -39,6 +45,12 @@ defmodule OnigumoCLITest do
...
+        assert usage_message_printed?(fn -> Onigumo.CLI.main([unquote(switches)]) end)
```
Glutexo, line 50:
```suggestion
        assert usage_message_printed?(fn -> Onigumo.CLI.main([unquote(combination)]) end)
```

test/onigumo_cli_test.exs
```diff
@@ -60,6 +72,12 @@ defmodule OnigumoCLITest do
       end
     end
 
+    for switch <- ["-h", "--help"] do
```
Glutexo, line 75: @help_switches ["-h", "--help"]

```suggestion
    for switch <- @help_switches do
```

🤔

test/onigumo_cli_test.exs
```diff
@@ -13,6 +13,12 @@ defmodule OnigumoCLITest do
     "-c"
   ]
 
+  @invalid_combinations [
+    "--help -C invalid",
+    "-h --invalid",
+    "downloader -h"
+  ]
+
```
Glutexo, line 21:
```suggestion

  @help_switches [
    "-h", "--help"
  ]
```

test/onigumo_cli_test.exs
```diff
@@ -60,6 +72,12 @@ defmodule OnigumoCLITest do
       end
     end
 
+    for switch <- ["-h", "--help"] do
+      test("run CLI with a #{inspect(switch)} switch") do
```
Glutexo, line 76:
```suggestion
      test("run CLI with #{inspect(switch)}") do
```

Consistency. 🤷🏻‍♂️

test/onigumo_cli_test.exs
```diff
@@ -39,6 +50,12 @@ defmodule OnigumoCLITest do
       end
     end
 
+    for combination <- @invalid_combinations do
+      test("run CLI with invalid combinations #{inspect(combination)} ") do
+        assert usage_message_printed?(fn -> Onigumo.CLI.main([unquote(combination)]) end)
```
Glutexo, line 55:
```suggestion
        assert usage_message_printed?(fn -> Onigumo.CLI.main(unquote(combination)) end)
```

test/onigumo_cli_test.exs
```diff
@@ -13,11 +13,22 @@ defmodule OnigumoCLITest do
     "-c"
   ]
 
+  @invalid_combinations [
+    "--help -C invalid",
+    "-h --invalid",
+    "downloader -h"
+  ]
```
Glutexo, line 20:
```suggestion
  @invalid_combinations [
    ["--help", "-C", "invalid"],
    ["-h", "--invalid"],
    ["downloader", "-h"]
  ]
```

`argv` is passed as a list of arguments, not as a space-separated string. As a result, you will see the `"downloader -h"` test failing. See my comments in https://github.com/Glutexo/onigumo/pull/238#issuecomment-2308952126.

Glutexo (changes requested):

Glutexo:
I created #260 that only fixes the invalid combinations. In `master`, however, it is not possible to have an invalid combinations. Because of that, there are no automated tests in #260. It can be tested manually though. There are tests instructions and a patch in the comments.

After merging #260, the change can be removed from this pull request. It can be then updated and re-reviewed.  Asking for a review of #260.

test/onigumo_cli_test.exs
```diff
@@ -13,11 +13,22 @@ defmodule OnigumoCLITest do
     "-c"
   ]
 
+  @invalid_combinations [
+    ["--help", "-C", "invalid"],
+    ["-h", "--invalid"],
```
Glutexo, line 18: This is, in fact, a different use case. It tests another branch of code: the one for invalid switches. The other combinations test a combination of a positional argument and a valid switch.

This exact case would work on _master_ too and should be tested. Not in this pull request though.

Glutexo: Created #261.

lib/cli.ex
```diff
@@ -36,6 +48,7 @@ defmodule Onigumo.CLI do
     COMPONENT\tOnigumo component to run, available: #{components}
 
     OPTIONS:
+    -h, --help\t\tPrint this help
```
Glutexo, line 51: Why two tabs? We should not assume tab size. Maybe we shouldn’t use tabs at all. For now, let’s use `\t` as in `-C`.

Created #265.

```suggestion
    -h, --help\tPrint this help
```

Glutexo (changes requested):

lib/cli.ex
nappex: Why to remove double `TAB`?

test/onigumo_cli_test.exs
```diff
@@ -13,11 +13,22 @@ defmodule OnigumoCLITest do
     "-c"
   ]
 
+  @invalid_combinations [
+    ["--help", "-C", ".", "downloader"],
+    ["downloader", "-h"],
+    ["downloader", "--help"]
```
Glutexo, line 19: This may be a little redundant, but ok.

test/onigumo_cli_test.exs
```diff
@@ -39,6 +50,12 @@ defmodule OnigumoCLITest do
       end
     end
 
+    for combination <- @invalid_combinations do
+      test("run CLI with invalid combination #{inspect(combination)} ") do
```
Glutexo, line 54:
```suggestion
      test("run CLI with invalid combination #{inspect(combination)}") do
```

Glutexo (changes requested):
Just fix the trailing space please.

PR #237, opened 2024-08-03
Update mermaid graph of onigumo processing with a new approach

nappex:

PR #239, opened 2024-08-16
Update description to match diagrams

Glutexo:
The Operator-Downloader-Parser trinity became only a part of the whole architecture. It’s no longer true that Onigumo is composed of only three components. It wasn’t entirely true before either: there was still the idea of a Materializer.

There are two flowcharts now. It’d be possible to call them a single one consisting of two parts, but clarity doesn’t harm.

Follow-up to #237.


nappex (approved):
really fast fix! Thank you

PR #242, opened 2024-08-25
Split cli messages to case with invalid usage and help message

### LINKED ISSUE #240: Split usage_message and help_message
We'd like to split output to user if he invokes full content of help content with help option `-h/--help` and message when user misused the the command onigumo with invalid usage. Breifly said we want to apply same approach as `curl` or `wget` does.

nappex:
fix #240 

One would be help message

another perhaps usage message for invalid usage of onifumo

This PR is done above the PR #238, because it is nature continuation of it.

Before merging this to master must be ensured merge of #238 to #242

### DIFF
--- a/lib/cli.ex
+++ b/lib/cli.ex
@@ -23,8 +23,6 @@ defmodule Onigumo.CLI do
   end
 
   defp usage_message() do
-    components = Enum.join(Map.keys(@components), ", ")
-
     IO.puts("""
     onigumo: invalid usage
     Usage: onigumo [OPTION]... [COMPONENT]
--- a/lib/cli.ex
+++ b/lib/cli.ex
@@ -37,4 +37,20 @@ defmodule Onigumo.CLI do
     -C, --working-dir <dir>\tChange working dir to <dir> before running
     """)
   end
+
+  defp help_message() do
+    components = Enum.join(Map.keys(@components), ", ")
+
+    IO.puts("""
+    Usage: onigumo [OPTION]... [COMPONENT]
+
+    Simple program that retrieves HTTP web content as structured data.
+
+    COMPONENT\tOnigumo component to run, available: #{components}
+
+    OPTIONS:
+    -h, --help\t\tprint this help
+    -C, --working-dir <dir>\tChange working dir to <dir> before running
+    """)
+  end
 end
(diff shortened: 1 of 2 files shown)


nappex:
Revise tests! 😈

nappex:
--help must be used as only one cli argument to run correctly, we do not want behaviour as "rule them all" by help switch as proposed in PR #244 .
If someone use --help with other argument or switch, then it could evoke to user it should output help for certain argument resp. switch. We do not want to implement this feature currently, maybe later. So --help should be used only for general help message for whole Onigumo not for its individual commands.

Glutexo:
The `-h`/`--help` is introduced in a separate pull request #238. This pull request should only introduce the help message.

lib/cli.ex
```diff
@@ -38,6 +38,15 @@ defmodule Onigumo.CLI do
   end
 
   defp usage_message() do
+    IO.puts("""
+    onigumo: invalid usage
+    Usage: onigumo [OPTION]... [COMPONENT]
```
Glutexo, line 43:
```suggestion
    Usage: onigumo [OPTION]… [COMPONENT]
```

What about Unicode elipsis?

test/onigumo_cli_test.exs
```diff
@@ -79,13 +79,18 @@ defmodule OnigumoCLITest do
...
 
     defp usage_message_printed?(function) do
       output = capture_io(function)
-      String.starts_with?(output, "Usage: onigumo ")
+      String.starts_with?(output, "onigumo: invalid usage")
+    end
+
+    defp help_message_printed?(function) do
+      output = capture_io(function)
+      String.starts_with?(output, "Usage: onigumo [OPTION]... [COMPONENT]")
```
Glutexo, line 93:
```suggestion
      String.starts_with?(output, "Usage: onigumo [OPTION]… [COMPONENT]")
```

lib/cli.ex
```diff
@@ -38,6 +38,15 @@ defmodule Onigumo.CLI do
...
+
+    Try `onigumo --help' for more options.
+    """)
```
Glutexo, line 46: This produces an excessive newline. `IO.puts` puts (😃) a `\n` at the end automatically. It happens for the usage message too though. Not in scope of this PR. Created https://github.com/Glutexo/onigumo/issues/266.

Glutexo (approved):
Awesome! ❤️

LGTM

PR #245, opened 2024-11-09
Replace case with with

Glutexo:
The `case` statement is too complex for situations when we want to focus on the success branch. This is the case of `OptionParser.parse`, which fails on invalid input, but requires complex parsing of correct switches and arguments.

### DIFF
--- a/lib/cli.ex
+++ b/lib/cli.ex
@@ -4,18 +4,15 @@ defmodule Onigumo.CLI do
   }
 
   def main(argv) do
-    case OptionParser.parse(
-           argv,
-           aliases: [C: :working_dir],
-           strict: [working_dir: :string]
-         ) do
-      {parsed_switches, [component], []} ->
-        {:ok, module} = Map.fetch(@components, String.to_atom(component))
-        working_dir = Keyword.get(parsed_switches, :working_dir, File.cwd!())
-        module.main(working_dir)
+    parse_result =
+      OptionParser.parse(argv, aliases: [C: :working_dir], strict: [working_dir: :string])
 
-      _ ->
-        usage_message()
+    with {parsed_switches, [component], []} <- parse_result do
+      {:ok, module} = Map.fetch(@components, String.to_atom(component))
+      working_dir = Keyword.get(parsed_switches, :working_dir, File.cwd!())
+      module.main(working_dir)
+    else
+      _ -> usage_message()
     end
   end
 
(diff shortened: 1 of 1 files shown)


lib/cli.ex
```diff
@@ -4,18 +4,15 @@ defmodule Onigumo.CLI do
...
-    case OptionParser.parse(
-           argv,
-           aliases: [C: :working_dir],
-           strict: [working_dir: :string]
-         ) do
-      {parsed_switches, [component], []} ->
-        {:ok, module} = Map.fetch(@components, String.to_atom(component))
-        working_dir = Keyword.get(parsed_switches, :working_dir, File.cwd!())
-        module.main(working_dir)
+    parse_result =
```
nappex, line 7: `parse` should be enough, but at least `parsed_result` with **d** is more semantically right in my opinion.

lib/cli.ex
```diff
@@ -4,18 +4,15 @@ defmodule Onigumo.CLI do
...
+      OptionParser.parse(argv, aliases: [C: :working_dir], strict: [working_dir: :string])
 
-      _ ->
-        usage_message()
+    with {parsed_switches, [component], []} <- parse_result do
```
nappex, line 10: There you can see that you call parsed_switches as `parsed_switches` but the result is not `parsed` only `parse`, it is not consistent.

lib/cli.ex
Glutexo: 🤔  `parsed_result` doesn’t seem right to me: it’s not a result that has been parsed; it’s a result of parsing. `parse_result` tried to convey the message that it’s a result of the `parse` function.

Shortened to a single word, `parsed` may be a better choice. `parse` sounds like an action rather than a value.

lib/cli.ex
Glutexo: `parse_result` is a result of a `parse` function. `parsed_switches` on the other hand are really switches that have been parsed. The “result” has not been parsed – the args were and the “result” is the result of that.

I don’t really like the `parsed_switches` name. But since it’s there, I think the most consistent option for `parse_result` would be just `parsed` or `parsed_argv`.

nappex:
Resources for finalize this PR:
- https://dev.to/martinthenth/using-elixirs-with-statement-5e36
- https://www.openmymind.net/Elixirs-With-Statement/
- https://ayoushchourasia.medium.com/demystifying-elixir-basics-a-comprehensive-guide-to-the-with-expression-part-3-of-our-elixir-93c9567701ac

nappex:
`case` was replaced by #250

PR #249, opened 2024-11-09
Help message on invalid component

Glutexo:
Passing an invalid component name prints a friendly help message instead of crashing with a stack trace. This is now consistent with a situation when an invalid switch is passed.

Fixes #189.

### DIFF
--- a/lib/cli.ex
+++ b/lib/cli.ex
@@ -10,9 +10,12 @@ defmodule Onigumo.CLI do
            strict: [working_dir: :string]
          ) do
       {switches, [component], []} ->
-        {:ok, module} = Map.fetch(@components, String.to_atom(component))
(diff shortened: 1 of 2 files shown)


Glutexo:
Shouldn’t I extend the with block to the `Keyword.get` call?

```elixir
with {:ok, module} <-
  Map.fetch(@components, String.to_atom(component)),
  working_dir <- Keyword.get(switches, :working_dir, File.cwd!())
```

But `Keyword.get` never fails and thus doesn’t utilize the `else` block. Another option would be to use `Keyword.fetch` instead, but it doesn’t support a default value. It would utilize the `else` block, but there, the default value would need to be assigned.

nappex (approved):
Nice 👍

PR #250, opened 2024-11-09
Remove redundant call of usage message

nappex:
DRY if it is not necessary, then dont call function more than once.

### DIFF
--- a/lib/cli.ex
+++ b/lib/cli.ex
@@ -4,21 +4,14 @@ defmodule Onigumo.CLI do
   }
 
   def main(argv) do
-    case OptionParser.parse(
-           argv,
-           aliases: [C: :working_dir],
-           strict: [working_dir: :string]
-         ) do
-      {switches, [component], []} ->
-        with {:ok, module} <- Map.fetch(@components, String.to_atom(component)) do
-          working_dir = Keyword.get(switches, :working_dir, File.cwd!())
-          module.main(working_dir)
-        else
-          :error -> usage_message()
-        end
+    parsed = OptionParser.parse(argv, aliases: [C: :working_dir], strict: [working_dir: :string])
 
-      _ ->
-        usage_message()
+    with {switches, [component], []} <- parsed,
+         {:ok, module} <- Map.fetch(@components, String.to_atom(component)) do
+      working_dir = Keyword.get(switches, :working_dir, File.cwd!())
+      module.main(working_dir)
+    else
+      _ -> usage_message()
     end
   end
 


lib/cli.ex
```diff
@@ -4,21 +4,14 @@ defmodule Onigumo.CLI do
...
+    parsed = OptionParser.parse(argv, aliases: [C: :working_dir], strict: [working_dir: :string])
 
-      _ ->
-        usage_message()
+    with {switches, [component], []} <- parsed,
+         {:ok, module} <- Map.fetch(@components, String.to_atom(component)) do
+      working_dir = Keyword.get(switches, :working_dir, File.cwd!())
+      module.main(working_dir)
+    else
+      _ -> usage_message()
```
nappex, line 14: maybe there could be `:error -> usage_message()` this should be check before approved

Glutexo: We can check for `{_, _, [_, _]}`, which would match if `OptionParse.parse` finds an invalid switch. The `[_, _]` matches a non-empty `errors` list.

The above, however, doesn’t match invalid argument count. For that, we’d need other matches for `argv`:

* `{_, [], []}` for no component;
* `{_, [_ | [_ | _], []} for mor than one component.

The latter (or both) may be solved more elegantly by a guard.

Such patterns would describe all expected (although erroneous) cases. If any real unexpected error would occur, it would be raised and not silently comsumed and reditrected to the usage message.

We can discuss on whether we always want to simply print the usage message. The more sophisticated mathing allows for more precise messages.

lib/cli.ex
```diff
@@ -4,21 +4,14 @@ defmodule Onigumo.CLI do
...
-          working_dir = Keyword.get(switches, :working_dir, File.cwd!())
-          module.main(working_dir)
-        else
-          :error -> usage_message()
-        end
+    parsed = OptionParser.parse(argv, aliases: [C: :working_dir], strict: [working_dir: :string])
 
-      _ ->
-        usage_message()
+    with {switches, [component], []} <- parsed,
```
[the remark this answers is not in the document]
nappex: I'd like to keep in a new variable for better legibility

lib/cli.ex
Glutexo: I explored the guards. We can leverage the `length` guard to check for the right size of the list.

* `{_, [], errors} when length(errors) > 0` for the switches. This is however discouraged by Elixir and doing so issues a warning. Apparently `[_ | _]` is the preferred way to check for a non-empty list. A descriptive alternative using a guard is `when errors != []`.
* `{_, argv, _} when length(argv) != 1` for the arguments.

Doing so nicely describes the intent. Comments or macros are thus not needed.

lib/cli.ex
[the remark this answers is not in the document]
Glutexo: 🤷🏻‍♂️

lib/cli.ex
```diff
@@ -4,21 +4,17 @@ defmodule Onigumo.CLI do
...
+         {:ok, module} <- Map.fetch(@components, String.to_atom(component)) do
+      working_dir = Keyword.get(switches, :working_dir, File.cwd!())
+      module.main(working_dir)
+    else
+      :error -> usage_message()
```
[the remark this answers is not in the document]
nappex: @Glutexo I had same idea today :D :D, yes it just a detail but could be nice

lib/cli.ex
```diff
@@ -4,21 +4,17 @@ defmodule Onigumo.CLI do
...
+      {_, _, [_ | _]} -> usage_message()
+      {_, argv, _} when length(argv) != 1 -> usage_message()
+      _ -> usage_message()
```
nappex, line 17: If we handle all cases then the not defined may not be catch

lib/cli.ex
```diff
@@ -4,21 +4,17 @@ defmodule Onigumo.CLI do
...
-        usage_message()
+    with {switches, [component], []} <- parsed,
+         {:ok, module} <- Map.fetch(@components, String.to_atom(component)) do
+      working_dir = Keyword.get(switches, :working_dir, File.cwd!())
+      module.main(working_dir)
+    else
+      {_, _, [_ | _]} -> usage_message()
+      {_, argv, _} when length(argv) != 1 -> usage_message()
+      :error -> usage_message()
+      _ -> usage_message()
```
Glutexo, line 17: Now, all cases are covered by the explicit patterns

* `{_, _, [_ | _]}` (invalid switches),
* `{_, argv, _} (invalid argument number)`,
* and `:error` (invalid component). 

Universal `_` would silently swallow any unexpected value. In that case we rather want an error to be raised.

Glutexo (changes requested):

PR #257, opened 2025-01-04
Another example with case and implementation of help

nappex:


nappex:
It was only example of two approaches: the `case` and the `with` to compare them and do not forget ideas. Finally we choose the solution from PR #256 and this example is not necessary to hold as active PR.

PR #260, opened 2025-03-20
Valid switches with args didn’t fail

Glutexo:
`Keyword.get :working_dir` doesn’t check for other switches. In the future, there may be more switches. Some of those switches may be mutually exclusive with a component name. But now, such switches would be ignored.

Imagine a help switch (`-h`/`--help`). This switch is only valid alone like `onigumo -h`. `onigumo -h downloader` or `onigumo --working-dir`. `downloader --help` are not valid. `Keyword.get` only checks for `:working_dir` though; `:help` is ignored.

`Keyword.pop` yields a keyword list of remaining switches. `:working_dir` is the only valid switch for a component. Any remaining switches mean invalid usage.

There is no way to test this: `-C` is the only valid switch. Mocking OptionParse would be an overkill: `-h` will be added soon. Then, this change can be tested.

Manual testing can be done:

1. Add `h: :help` to `OptionParser.parse` `:aliases` and `help: boolean` to `:strict`.
2. Recompile (mix escript.build) and run `./onigumo -h downloader`.
3. See the usage message printed. Downloader should not be run.

Blocks #238.

### DIFF
--- a/lib/cli.ex
+++ b/lib/cli.ex
@@ -8,8 +8,12 @@ defmodule Onigumo.CLI do
 
     with {switches, [component], []} <- parsed,
          {:ok, module} <- Map.fetch(@components, String.to_atom(component)) do
-      working_dir = Keyword.get(switches, :working_dir, File.cwd!())
-      module.main(working_dir)
+      {working_dir, switches} = Keyword.pop(switches, :working_dir, File.cwd!())
+
+      case switches do
+        [] -> module.main(working_dir)
+        _ -> usage_message()
+      end
     else
       {_, _, [_ | _]} -> usage_message()
       {_, argv, _} when length(argv) != 1 -> usage_message()
--- a/lib/cli.ex
+++ b/lib/cli.ex
@@ -16,6 +16,7 @@ defmodule Onigumo.CLI do
               [] -> module.main(working_dir)
               _ -> usage_message()
             end
+
           :error ->
             usage_message()
         end


Glutexo:
This can tested by applying the patch from https://github.com/Glutexo/onigumo/tree/pop-patch`.

Glutexo:
I didn’t push the merge with [master](https://github.com/Glutexo/onigumo/tree/master). Fixed.

Glutexo:
The failing CI is not caused by the changes in this pull request. See #262, which fixes the problem.

Glutexo:
https://github.com/Glutexo/onigumo/pull/262 has been merged. Merging this to restart the CI with the new settings.

Glutexo:
Fixed formatting, CI passed. :shipit:

nappex (approved):
The code is OK.
But I prefer merge this code to my PR #238, where we are testing that usage message is printed when --help is call with other arguments or switches as invalid combination `["downloader", "-h"]`. The reason why I want to merge it is because it solve the problem of the PR. Finally merge these two PR as one with master then.

Glutexo:
@nappex We can try what happens. You can merge [pop](https://github.com/Glutexo/onigumo/tree/pop) into #238. Merging your pull request will then merge both branches into _master_. That should close this pull request automatically then. Let’s see what GitHub does.

Glutexo:
Got merged! Woohoo!

PR #274, opened 2025-04-11
Run CI with os matrix

### LINKED ISSUE #269: Add Windows to CI matrix
We are not sure how Elixir handles newlines.

nappex:
fix #269

### DIFF
--- a/.github/workflows/elixir.yml
+++ b/.github/workflows/elixir.yml
@@ -33,4 +33,5 @@ jobs:
     - name: Run tests
       run: mix test
     - name: Check formatting
+      if: startsWith(matrix.os, 'ubuntu')
       run: mix format --check-formatted
--- a/.github/workflows/elixir.yml
+++ b/.github/workflows/elixir.yml
@@ -9,7 +9,7 @@ on:
 jobs:
   build:
 
-    name: ${{matrix.os}} - Build and test
+    name: ${{matrix.os}} – Build and test
     runs-on: ${{matrix.os}}
     strategy:
       matrix:
--- a/.github/workflows/elixir.yml
+++ b/.github/workflows/elixir.yml
@@ -9,8 +9,11 @@ on:
 jobs:
   build:
 
-    name: Build and test
-    runs-on: ubuntu-24.04
+    name: ${{matrix.os}} - Build and test
+    runs-on: ${{matrix.os}}
+    strategy:
+      matrix:
+        os: [ubuntu-24.04, windows-2022]
 
     steps:
     - name: Check out the repository
--- a/.github/workflows/elixir.yml
+++ b/.github/workflows/elixir.yml
@@ -37,5 +37,4 @@ jobs:
     - name: Run tests
       run: mix test
     - name: Check formatting
-      if: startsWith(matrix.os, 'ubuntu')
       run: mix format --check-formatted


nappex:
Windows are failling during `--check-formatted` I supposed we do not need to pass this check on windows.
What do you think. Currently I am skipping with condition in github action.

Glutexo:
I’d like to see why it’s failing. The condition is a good idea. We can fix the Windows issue separately and then remove the condition.

Thank you tor the pull request! We need Windows to see how it deals with `\r\n` newlines.

Glutexo:
I’d like to see why it’s failing. The condition is a good idea. We can fix the Windows issue separately and then remove the condition.

Thank you tor the pull request! We need Windows to see how it deals with `\r\n` newlines.

.github/workflows/elixir.yml
```diff
@@ -30,4 +33,5 @@ jobs:
     - name: Run tests
       run: mix test
     - name: Check formatting
+      if: startsWith(matrix.os, 'ubuntu')
```
Glutexo, line 36:
```suggestion
      if: startsWith(matrix.os, 'ubuntu-')
```

Glutexo:
I figured out that setting [`core.autocrlf`](https://git-scm.com/docs/git-config#Documentation/git-config.txt-coreautocrlf) to `false` does the trick. See https://github.com/Glutexo/onigumo/pull/276. Apparently, the default value is `true` on `windows-2022`. What does the configuration do?

> Setting this variable to "true" is the same as setting the text attribute to "auto" on all files and core.eol to "crlf". Set to true if you want to have CRLF line endings in your working directory and the repository has LF line endings.

That means that Git converts the line endings to `\r\n`. Even though we have `\n` in the repository. And `mix format` [requires](https://hexdocs.pm/elixir/Code.html#format_string!/2-newlines) `\n`, it fails on `\r\n`:

> The formatter converts all newlines in code from \r\n to \n.

---

More on the [`core.autocrlf`](https://git-scm.com/docs/git-config#Documentation/git-config.txt-coreautocrlf) option: it overrides [`core.eol`](https://git-scm.com/docs/git-config#Documentation/git-config.txt-coreeol) to `crlf`. Setting `core.eol` thus doesn’t help. Setting `.gitattributes` wouldn’t help either: the config option overrides those too.

I’d like to see, however, where it’s documented that the default for `core.autocrlf` is `true`. I am not sure whether that applies to all Windows, or to the GitHub actions OS image.

Glutexo:
#276 has been merged. The `mix format` condition is no longer necessary. It will work on Windows, too.

.github/workflows/elixir.yml
```diff
@@ -9,8 +9,11 @@ on:
 jobs:
   build:
 
-    name: Build and test
-    runs-on: ubuntu-24.04
+    name: ${{matrix.os}} - Build and test
```
Glutexo, line 12:
```suggestion
    name: ${{matrix.os}} – Build and test
```

.github/workflows/elixir.yml
```diff
@@ -9,8 +9,11 @@ on:
 jobs:
   build:
 
-    name: Build and test
-    runs-on: ubuntu-24.04
+    name: ${{matrix.os}} – Build and test
```
Glutexo, line 12: Much nicer!

PR #275, opened 2025-04-11
Usg msg reason

### LINKED ISSUE #251: Add option to `usage_message` to write certain reason of problem to output
```elixir
usage_message("too many arguments")

defp usage_message(reason="") do
    components = Enum.join(Map.keys(@components), ", ")
    if reason:
         reason = "Invalid usage: #{reason}

    IO.puts("""
    #{reason}
    Usage: onigumo [OPTION]... [COMPONENT]

    OPTIONS:
    -C, --working-dir <dir>\tChange working dir to <dir> before running
    """)
  end
```

- this approach should improve also behavior of tests

nappex:
fix #251

### DIFF
--- a/lib/cli.ex
+++ b/lib/cli.ex
@@ -37,14 +37,20 @@ defmodule Onigumo.CLI do
     end
   end
 
-  defp usage_message() do
+  defp usage_message(reason \\ nil) do
     IO.write("""
-    onigumo: invalid usage
+    onigumo: invalid usage#{reason_message(reason)}
     Usage: onigumo [OPTION]... [COMPONENT]
 
     Try `onigumo --help' for more options.
     """)
   end
+
+  defp reason_message(reason) when is_nil(reason) or byte_size(reason) == 0 do
+    ""
+  end
+
+  defp reason_message(reason), do: ", #{reason}"
 
   defp help_message() do
     components = Enum.join(Map.keys(@components), ", ")
--- a/lib/cli.ex
+++ b/lib/cli.ex
@@ -22,18 +22,18 @@ defmodule Onigumo.CLI do
 
             case switches do
               [] -> module.main(working_dir)
-              _ -> usage_message()
+              _ -> usage_message("too many options")
             end
 
           :error ->
-            usage_message()
+            usage_message("missing value for working directory")
         end
 
       {_, _, [_ | _]} ->
-        usage_message()
+        usage_message("unknown options")
 
       {_, argv, _} when length(argv) != 1 ->
-        usage_message()
+        usage_message("exact one positional argument must be used")
     end
   end
 


lib/cli.ex
```diff
@@ -22,30 +22,33 @@ defmodule Onigumo.CLI do
...
-              _ -> usage_message()
+              _ -> usage_message("unknown options")
             end
 
           :error ->
-            usage_message()
+            usage_message("unknown positional argument")
         end
 
       {_, _, [_ | _]} ->
```
Glutexo, line 32: To be specific below, we’d need to name the third element.

```suggestion
      {_, _, invalid} when length(invalid) > 0 ->
```

lib/cli.ex
```diff
@@ -22,30 +22,33 @@ defmodule Onigumo.CLI do
...
+    onigumo: invalid usage#{reason_message(String.trim(reason))}
     Usage: onigumo [OPTION]... [COMPONENT]
 
     Try `onigumo --help' for more options.
     """)
   end
 
+  defp reason_message(reason) when byte_size(reason) == 0, do: reason
+  defp reason_message(reason), do: ", #{reason}"
+
```
Glutexo, line 51: This may be too complicated. If we make sure there is always a reason, we don’t need this function.

```suggestion
```

Glutexo (changes requested):
Thanks for the pull request! ❤️ 

Comments marked as ❗️ are blocking. The other ones are more comments and suggestions.

lib/cli.ex
```diff
@@ -22,30 +22,33 @@ defmodule Onigumo.CLI do
...
     IO.write("""
-    onigumo: invalid usage
+    onigumo: invalid usage#{reason_message(String.trim(reason))}
     Usage: onigumo [OPTION]... [COMPONENT]
 
     Try `onigumo --help' for more options.
     """)
   end
 
+  defp reason_message(reason) when byte_size(reason) == 0, do: reason
```
Glutexo, line 49: If `nil` is used for no message instead of `""`, the guard can be simpler.

```suggestion
  defp reason_message(reason) when is_nil(reason), do: reason
```

If I am not mistaken, however, there doesn’t even need to be a guard.

```suggestion
  defp reason_message(""), do: reason
```

```suggestion
  defp reason_message(nil), do: reason
```

lib/cli.ex
```diff
@@ -22,30 +22,33 @@ defmodule Onigumo.CLI do
 
             case switches do
               [] -> module.main(working_dir)
-              _ -> usage_message()
+              _ -> usage_message("unknown options")
```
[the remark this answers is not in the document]
nappex: then the message will be `: invalid usage, invalid options` that sounded strange to me.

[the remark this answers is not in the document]
nappex: Maybe we do not have to use `Enum` the `Inspect should be fine to show the list`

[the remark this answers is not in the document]
Glutexo: `inspect` is however mainly a debugging function. It exposes internal information about the Elixir objects.

It also puts escape sequences into the strings. We don’t expect options to contain e.g. `\n`, but it shows that it’s not a simple print.

lib/cli.ex
```diff
@@ -22,24 +22,24 @@ defmodule Onigumo.CLI do
 
             case switches do
               [] -> module.main(working_dir)
-              _ -> usage_message()
+              _ -> usage_message("invalid OPTIONS")
```
Glutexo, line 25:
```suggestion
              _ -> usage_message("incompatible OPTIONS")
```

in case you don’t like „invalid”.

---

It looks like there is no way to get the exact switch literal used in _argv_. We can’t discern whether the user provided `-h` or `--help`.

It may, however, not be a problem. The help message always lists a long version along with its short alias. Each switch has the long version.

I mean, printing “invalid option help” or “invalid option --help” may not be that bad even if the user provided `-h`. It may look slightly confusing, but at the same time it provides useful additional information.

There is a function `OptionParser.to_argv/1` that translates argument names and values to their CLI counterparts. It can take the `invalid` keyword list returned by `OptionParser.parse/3`.

```elixir
iex(1)> OptionParser.to_argv([help: true])
["--help"]
```

nappex: thank you I think `OptionParser.to_argv` will be good choice to both `help` and `invalid`

lib/cli.ex
nappex:
```Compiling 1 file (.ex)
    warning: do not use "length(invalid) > 0" to check if a list is not empty since length always traverses the whole list. Prefer to pattern match on a non-empty list, such as [_ | _], or use "invalid != []" as a guard```

File: lib/cli.ex (hunk 5 above)
Comment (nappex) [reply lib/cli.ex:25]: Ok it seems it cannot be used for `invalid`

File: lib/cli.ex (hunk 4 above)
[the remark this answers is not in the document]
Comment (nappex) [reply lib/cli.ex]: `Keyword.keys(switches)` can not be used to `invalid` because `invalid` is `List` with `Tuples` with two `Elements`, where `Elements` are `String` or `nil`. `Keyword` must be the first `Element` the `Atom`

File: lib/cli.ex [hunk 6]
```diff
@@ -22,24 +22,26 @@ defmodule Onigumo.CLI do
 
             case switches do
               [] -> module.main(working_dir)
-              _ -> usage_message()
+              _ -> usage_message("invalid OPTIONS #{inspect(OptionParser.to_argv(switches))}")
             end
 
           :error ->
-            usage_message()
+            usage_message("invalid COMPONENT #{inspect(component)}")
```
Glutexo, line 29:
```suggestion
            usage_message("invalid COMPONENT #{component}")
```

lib/cli.ex
```diff
@@ -22,24 +22,26 @@ defmodule Onigumo.CLI do
...
         end
 
-      {_, _, [_ | _]} ->
-        usage_message()
+      {_, _, invalid} when invalid != [] ->
```
Glutexo, line 32: I found out that the guard can be omitted.

```suggestion
      {_, _, invalid = [_ | _]} ->
```

keeps the original pattern, only attaching a name to it.

lib/cli.ex
```diff
@@ -21,25 +21,32 @@ defmodule Onigumo.CLI do
...
 
             case switches do
-              [] -> module.main(working_dir)
-              _ -> usage_message()
+              [] ->
+                module.main(working_dir)
+
+              _ ->
+                usage_message(
+                  "invalid OPTIONS #{Enum.join(OptionParser.to_argv(switches), ", ")}"
```
Glutexo, line 29:
```suggestion
                  options = Enum.join(OptionParser.to_argv(switches)
                  "invalid OPTIONS #{Enum.join(OptionParser.to_argv(switches), ", ")}"
```

```suggestion
                  "invalid OPTIONS #{switches |> OptionParser.to_argv() |> Enum.join(", ")}"
```

lib/cli.ex
```diff
@@ -21,25 +21,32 @@ defmodule Onigumo.CLI do
...
 
           :error ->
-            usage_message()
+            usage_message("invalid COMPONENT #{component}")
         end
 
-      {_, _, [_ | _]} ->
-        usage_message()
+      {_, _, invalid = [_ | _]} ->
+        Enum.map(invalid, fn {k, _} -> k end)
```
Glutexo, line 38:
```suggestion
        Enum.map(invalid, &at(&1, 0))
```

lib/cli.ex
```diff
@@ -21,25 +21,32 @@ defmodule Onigumo.CLI do
...
+        |> then(&"invalid OPTIONS #{Enum.join(&1, ", ")}")
+        |> usage_message()
 
       {_, argv, _} when length(argv) != 1 ->
```
Glutexo, line 42: This guard is probably unnecessary, I‘ll create an issue for that.

lib/cli.ex
```diff
@@ -21,25 +21,32 @@ defmodule Onigumo.CLI do
...
 
             case switches do
-              [] -> module.main(working_dir)
-              _ -> usage_message()
+              [] ->
+                module.main(working_dir)
+
+              _ ->
+                OptionParser.to_argv(switches)
+                |> then(&"invalid OPTIONS #{Enum.join(&1, ", ")}")
```
Glutexo, line 29: This can be DRIED to a private function accepting a list of invalid options in the form of argv.

```suggestion
invalid_options()
```

lib/cli.ex
```diff
@@ -21,25 +21,32 @@ defmodule Onigumo.CLI do
...
           :error ->
-            usage_message()
+            usage_message("invalid COMPONENT #{component}")
         end
 
-      {_, _, [_ | _]} ->
-        usage_message()
+      {_, _, invalid = [_ | _]} ->
+        Enum.map(invalid, &elem(&1, 0))
+        |> then(&"invalid OPTIONS #{Enum.join(&1, ", ")}")
```
Glutexo, line 39:
```suggestion
        |> invalid_options()
```

lib/cli.ex
```diff
@@ -21,25 +21,32 @@ defmodule Onigumo.CLI do
...
+        |> usage_message()
 
       {_, argv, _} when length(argv) != 1 ->
-        usage_message()
+        usage_message("exactly one COMPONENT must be provided")
     end
   end
 
-  defp usage_message() do
```
Glutexo, line 40:
```suggestion
defp invalid_options(options) do
  "invalid OPTIONS – #{Enum.join(options)}"
end

defp usage_message(reason) do
```

lib/cli.ex
```diff
@@ -21,25 +21,32 @@ defmodule Onigumo.CLI do
...
+
+              _ ->
+                OptionParser.to_argv(switches)
+                |> then(&"invalid OPTIONS #{Enum.join(&1, ", ")}")
+                |> usage_message()
             end
 
           :error ->
-            usage_message()
+            usage_message("invalid COMPONENT #{component}")
```
Glutexo, line 34:
```suggestion
            invalid_component(component) |> usage_message()
```

test/onigumo_cli_test.exs
```diff
@@ -32,27 +33,43 @@ defmodule OnigumoCLITest do
...
+               )
       end
     end
 
     for combination <- @invalid_combinations do
       test("run CLI with invalid combination #{inspect(combination)}") do
-        assert usage_message_printed?(fn -> Onigumo.CLI.main(unquote(combination)) end)
+        assert usage_message_printed?(
+                 fn -> Onigumo.CLI.main(unquote(combination)) end,
+                 # there is no easy way to do it dynamically
```
Glutexo, line 70: Yes, this sounds fishy.

lib/cli.ex
Glutexo, line 29:
```suggestion
                |> then(&"incompatible OPTIONS #{Enum.join(&1, ", ")}")
```

test/onigumo_cli_test.exs
```diff
@@ -32,27 +33,43 @@ defmodule OnigumoCLITest do
...
+                 "incompatible OPTIONS --help"
```
Glutexo, line 71:
```suggestion
                 "incompatible OPTIONS #{invalid}"
```

If the constant contained tuples as imagined above, the string would look like this.

test/onigumo_cli_test.exs
```diff
@@ -96,9 +116,9 @@ defmodule OnigumoCLITest do
       String.match?(output, ~r/\N\n\z/)
     end
 
-    defp usage_message_printed?(function) do
+    defp usage_message_printed?(function, reason) do
       output = capture_io(function)
-      String.starts_with?(output, "onigumo: invalid usage")
+      String.starts_with?(output, "onigumo: invalid usage – #{reason}")
```
Glutexo, line 121: Just thinking out loud.

```suggestion
      message = "onigumo: invalid usage – #{reason}"
      String.starts_with?(output, message)
```

test/onigumo_cli_test.exs
```diff
@@ -9,8 +9,9 @@ defmodule OnigumoCLITest do
   ]
 
   @invalid_switches [
-    "--invalid",
-    "-c"
+    ["--invalid"],
+    ["-c"],
+    ["-a", "-b"]
```
Glutexo, line 14: 🤔  I am not sure about this. It adds the possibility to test joining for the usage message. But this test should have been there in the first place. Or at least list of invalid switches instead of a single value. I feel it doesn’t belong into this PR.

However, without the list support and this test in place, it is not possible to test the new feature.

Glutexo (changes requested):
I pick Request changes for only two reasons:

* The `# there is no easy way to do it dynamically` comment doesn’t feel appropriate. There is and easy way. Either do it, change the comment to be more descriptive, or omit it at all.
* The `@invalid_switches` extension deserves a preliminary PR that would unblock this one.

lib/cli.ex
nappex: OK

test/onigumo_cli_test.exs
```diff
@@ -32,27 +33,43 @@ defmodule OnigumoCLITest do
...
       test("run CLI with invalid switch #{inspect(switch)}") do
-        assert usage_message_printed?(fn -> Onigumo.CLI.main([unquote(switch)]) end)
+        assert usage_message_printed?(
+                 fn -> Onigumo.CLI.main(unquote(switch)) end,
+                 "invalid OPTIONS #{Enum.join(unquote(switch), ", ")}"
+               )
       end
     end
 
     for combination <- @invalid_combinations do
```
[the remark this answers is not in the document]
nappex: I am not sure if it is just one value, it feels like unnecessary DRY at this moment.

lib/cli.ex
```diff
@@ -21,25 +21,32 @@ defmodule Onigumo.CLI do
...
         end
 
-      {_, _, [_ | _]} ->
-        usage_message()
+      {_, _, invalid = [_ | _]} ->
+        Enum.map(invalid, &elem(&1, 0))
+        |> then(&"invalid OPTIONS #{Enum.join(&1, ", ")}")
```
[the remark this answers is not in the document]
nappex: Alrigth why not, it looks like more clear and legibility, thank you.

test/onigumo_cli_test.exs
nappex: yes it is more clean suppose

nappex: but then it is for another PR since the same way of solving is for help message

test/onigumo_cli_test.exs
nappex: I would keep it

test/onigumo_cli_test.exs
```diff
@@ -9,8 +9,9 @@ defmodule OnigumoCLITest do
   ]
 
   @invalid_switches [
-    "--invalid",
-    "-c"
+    ["--invalid"],
+    ["-c"],
```
Glutexo, line 13:
```suggestion
    "--invalid",
    "-c"
```

Glutexo (commented):
I‘ll approve this once the `@invalid_switches` matter is resolved. Either of the suggested approaches is fine.

lib/cli.ex
```diff
@@ -21,25 +21,34 @@ defmodule Onigumo.CLI do
...
             case switches do
-              [] -> module.main(working_dir)
-              _ -> usage_message()
+              [] ->
+                module.main(working_dir)
+
+              _ ->
+                OptionParser.to_argv(switches)
+                |> Enum.join(", ")
+                |> then(&"incompatible OPTIONS #{&1}")
```
Glutexo, line 30:
```suggestion
                |> Enum.at(0)
                |> then(&"incompatible OPTION #{&1}")
```

This would allow to use the current invalid switch test. That is, without support for multiple invalid switches.

Glutexo: Created #279 for a test update that would save us from this.

test/onigumo_cli_test.exs
```diff
@@ -32,27 +33,42 @@ defmodule OnigumoCLITest do
...
+               "exactly one COMPONENT must be provided"
+             )
     end
 
     for switch <- @invalid_switches do
       test("run CLI with invalid switch #{inspect(switch)}") do
-        assert usage_message_printed?(fn -> Onigumo.CLI.main([unquote(switch)]) end)
+        assert usage_message_printed?(
+                 fn -> Onigumo.CLI.main(unquote(switch)) end,
+                 "invalid OPTIONS #{Enum.join(unquote(switch), ", ")}"
```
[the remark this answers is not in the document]
Glutexo: Created #279 for a test update that would save us from this.

lib/cli.ex
Glutexo: Resolved by merging #279.

Glutexo (approved):
LGTM

PR #276, opened 2025-04-17
💚 GitHub action matrix

Glutexo:
The `windows-2022` image has [`core.autocrlf`](https://git-scm.com/docs/git-config#Documentation/git-config.txt-coreautocrlf) set to `true`. This Git configuration replaces line endings to `\r\n`. `mix format`, however, [expects](https://hexdocs.pm/elixir/Code.html#format_string!/2-newlines) `\n`. The check then fails.

Setting `core.autocrlf` to `false` makes Git leave the newlines intact. On `ubuntu-24.01`, the value is false by default. Re-setting it does no harm.

The newline replacement happens during the clone. The setting needs to happen before. It must be global: before the clone, there is no local repository to set it to.

---

Fixes the CI error in #274.

### DIFF
--- a/.github/workflows/elixir.yml
+++ b/.github/workflows/elixir.yml
@@ -14,7 +14,7 @@ jobs:
 
     steps:
     - name: Configure Git
-      run: git config core.eol lf
+      run: git config --global core.eol lf
     - name: Check out the repository
       uses: actions/checkout@v3
     - name: Set up Elixir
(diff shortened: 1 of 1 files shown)


Glutexo:
Closing in order to create a new one based on [nappex/github-action-matrix](https://github.com/nappex/onigumo/tree/github-action-matrix).

nappex (approved):

Issue #278, opened 2025-04-25
🔥 Remove unnecessary guard


Glutexo:
https://github.com/Glutexo/onigumo/blob/6aaae5c1e5bb3c80fb64393829bcea062acf2d66/lib/cli.ex#L35

The guard in the last `case` branch in `Onigumo.CLI.main` is redundand. 

---

https://github.com/Glutexo/onigumo/blob/6aaae5c1e5bb3c80fb64393829bcea062acf2d66/lib/cli.ex#L32

The non-empty `invalid` branch matches **all** cases when `invalid` is not `[]`. All other branches thus always match `[]` on the third position. Including the last one on L35. We can even replace make it explicit: `{_, argv, []}` Whatever goes past `{_, _, [_ | _]}`, has `[]` as the last element.

---

https://github.com/Glutexo/onigumo/blob/6aaae5c1e5bb3c80fb64393829bcea062acf2d66/lib/cli.ex#L18

Then, the switches-component branch matches **all** remaining cases when `argv` only contains one element. The `switches` variable matches anything. What doesn’t match this one with `[]` invalid list, matches the aforementioned one `{_, _, [_ | _]}`.

---

Together, those two cover **all** cases `when length(argv) == 1`.

* `{switches, [component], []}` for the empty `invalid` list,
* `{_, _, [_ | _]}` for the non-empty one.

---

The remaining cases are:

* no positional argument (`when length(argv) == 0`) – gets either matched by `[help: true]` or `[_ | _]`, otherwise it goes all the way down to `{_, argv, _}`.
* more than one positional argument (`when length(argv) > 1`) – can only get matched by `{_, _, [_ | _]}` in case of invalid switches, otherwise bubbles to `{_, argv, _}` too.

---

Ergo, the guard can be safely omitted. Moreover, the third element can be explicitly defined as `[]`.

nappex:
so the solution is `_ -> "invalid number of positional arguments"`?

Glutexo:
`{_, _, []}` nicely describes a know case defined by the other cases. `_` may hypothetically miscatch cases where `OptionParser.parse` returns something else (not a triplet), or our error in the code. In such case the usage message would be printed. An error should be raised instead.

PR #282, opened 2025-05-24
Remove guard from more arugments

Glutexo:
The length guard for arguments was redundant. The other case (exactly one positional argument) is laredy covered by previous case branches. The reasoning described in more detail in https://github.com/Glutexo/onigumo/issues/278.

Fixes https://github.com/Glutexo/onigumo/issues/278.


Glutexo:
> `[]}`?

I did a separate pull request for that, since I considered it to be a separate (although related) issue.

Glutexo:
Thanks for the approval! <3 Merging.

PR #283, opened 2025-05-25
Match empty invalid in CLI

Glutexo:
The more than one component error branch always matches an empty invalid list. Mentioning it explicitly to make the match precise: preventing an invalid switch value to be falsely masked as more than one argument error.

Related to #282.


Glutexo:
This is still ready for review. May I ask for one, @nappex? 🙏🏻

nappex (approved):

Glutexo:
Thanks! <3

